PR #8, opened 2018-06-21
Training with quantization

guyjacob:
Added the capability to train with quantization in the loop, required by nearly all low precision methods under 8-bits.


guyjacob:
Last 3 commits fix comments from code review. Updated docs to follow in another commit.

PR #16, opened 2018-07-04
Prevent display of ‘float_weight’, allow bias in textual sparsity table

rotx-maxim:


guyjacob:
The premise of both changes looks fine.
Re disregarding 'float_weight': Looking at the code, I think I'd rather change the condition that looks at ['weights', 'bias'] to be even more strict, so it looks just for **_parameters_** named as such, because as it is now it allows for weird cases where the word 'weight' is used in the module name, but the parameter is actually named something else. Something like that seems unlikely to happen, but still it's better that we be explicit about what we're looking for.
In the interest of saving time I'll make the change and push it myself.

Thanks for bringing this up!

guyjacob:
Fixed in #20, closing.

PR #17, opened 2018-07-04
Allow quantize_bias to work in DorefaQuantizer and checkpoints

rotx-maxim:


guyjacob:
Just to note that the original DoReFa implementation explicitly quantizes only weights (see [here](https://github.com/tensorpack/tensorpack/blob/master/examples/DoReFa-Net/alexnet-dorefa.py#L76)).
But you made the default to be False so I'm fine with that. No reason to prevent experimentation.

Thanks for the PR!

PR #18, opened 2018-07-04
Allow ‘device’ specification in file_config so we can run on CPU

rotx-maxim:


nzmora:
Thanks Robert!
I think you forgot to include other changes you've made, because this PR seems incomplete.
For example, this change:

-    schedule = distiller.CompressionScheduler(model)
+    schedule = distiller.CompressionScheduler(model, device=device)

It is not sufficient to pass the `device` to CompressionScheduler - that object needs to use this device and make sure that everything it create uses the device (e.g. pruning masks).  There are other object that need to use this device, etc.
For example, this line places the target buffer on the GPU:
         target = target.cuda(async=True)
(from https://github.com/NervanaSystems/distiller/blob/master/examples/classifier_compression/compress_classifier.py).
It's a good start - I would continue by searching for all instances of "cuda" in the code, and plan how to propagate the device knowledge throughout the library and application.
Cheers,
Neta

rotx-maxim:
Yes and no. I verified that this works on CPU and GPU for me, but I'm neither using your example program nor any other quantizer or functionality. I'm also using pytorch-master (side note: async=True went away in 0.4.0). I think the biggest changes would be in the example programs.
In my code, I pass around `device` to everything, but that's a much bigger change. For this PR, I wasn't as ambitious and tried to create a non-breaking change, and I thought I did.

nzmora:
Thanks - It's clearer now why you wrote this PR as you did.
Distiller's treatment of the execution-device is not good: on the one hand we are not device-agnostic, on the other hand, some of the code passes around the device as a parameter.  This latter part is especially problematic because people may get the wrong impression that the code is device-agnostic.
The current PR is a non-breaking change, but it can create more confusion for people who expect that once we configure CompressionScheduler with a specific device, the code will use that device exclusively.
Therefore, I'd rather not integrate this PR as it is.  We should add a complete support for device compute placement.

nzmora:
We finally added support for running on CPU, so closing this PR.
Thanks @rotx-maxim !

PR #25, opened 2018-07-06
Allow quantize_bias to work in WRPNQuantizer

rotx-maxim:
Modification similar to #17 to allow experimentation with WRPNQuantizer.


guyjacob:
Looks good, thanks!

PR #32, opened 2018-07-19
Earlyexit

haim-barad:
Enablement of Early Exit strategy into the Classification example code.


examples/classifier_compression/compress_classifier.py
```diff
@@ -321,14 +327,16 @@ def main():
...
+              lossweights=args.lossweights, dataset=args.dataset)
         distiller.log_weights_sparsity(model, epoch, loggers=[tflogger, pylogger])
         if args.activation_stats:
             distiller.log_activation_sparsity(epoch, loggers=[tflogger, pylogger],
                                               collector=activations_sparsity)
 
         # evaluate on validation set
-        top1, top5, vloss = validate(val_loader, model, criterion, [pylogger], args.print_freq, epoch)
+        top1, top5, vloss = validate(val_loader, model, criterion, [pylogger], args.print_freq,
+                                     args.earlyexit, epoch, dataset=args.dataset)
```
nzmora: I prefer that you add new function arguments to the end of the argument list and give them a default value that renders them ineffectual unless explicitly used.  This helps other people using our code.

examples/classifier_compression/compress_classifier.py
```diff
@@ -361,6 +369,10 @@ def train(train_loader, model, criterion, optimizer, epoch,
     classerr = tnt.ClassErrorMeter(accuracy=True, topk=(1, 5))
     batch_time = tnt.AverageValueMeter()
     data_time = tnt.AverageValueMeter()
+    if lossweights:
+        exit0err = tnt.ClassErrorMeter(accuracy=True, topk=(1, 5))
+        exit1err = tnt.ClassErrorMeter(accuracy=True, topk=(1, 5))
+        exitNerr = tnt.ClassErrorMeter(accuracy=True, topk=(1, 5))
```
nzmora: This should be:
(1) more explicit - instead of ```if lossweights``` do something like ```if early_exit```.  I think this makes the code more explicit in your intentions.
(2) Make things more generic.  The code is written for 3 exits only.  What if someone wants 4 exits?  Or 2?  
For example, you can change the original code from:
```
classerr = tnt.ClassErrorMeter(accuracy=True, topk=(1, 5))
```
To 
```
classerr = [tnt.ClassErrorMeter(accuracy=True, topk=(1, 5)) * num_exits]
```

examples/classifier_compression/compress_classifier.py
```diff
@@ -382,11 +394,28 @@ def train(train_loader, model, criterion, optimizer, epoch,
         # Execute the forward phase, compute the output and measure loss
         if compression_scheduler:
             compression_scheduler.on_minibatch_begin(epoch, train_step, steps_per_epoch, optimizer)
-        output = model(input_var)
-        loss = criterion(output, target_var)
 
-        # Measure accuracy and record loss
-        classerr.add(output.data, target)
+        if lossweights:
```
nzmora: See my previous comments.
Also, when ordering the true clauses of an 'if' statement, I prefer to place the most common case in the 'True clause' and the less common case in the 'False clause'.  So here:
```if not early_exit:
         bla bla bla
     else:
         other bla bla bla
```

examples/classifier_compression/compress_classifier.py
```diff
@@ -382,11 +394,28 @@ def train(train_loader, model, criterion, optimizer, epoch,
...
+            if dataset == 'cifar10':
+                loss = ((lossweights[0] * criterion(output[0], target_var)) + ((1.0 - lossweights[0]) * criterion(output[1], target_var)))
+                exit0err.add(output[0].data, target)
+                exitNerr.add(output[1].data, target)
+            else:             # imagenet
+                loss = ((lossweights[0] * criterion(output[0], target_var)) + (lossweights[1] * criterion(output[1], target_var)) +
+                        ((1.0 - (lossweights[0]+lossweights[1])) * criterion(output[2], target_var)))
+                exit0err.add(output[0].data, target)
+                exit1err.add(output[1].data, target)
+                exitNerr.add(output[2].data, target)
```
nzmora: Same comment as before regarding the generalizing the code.
Also, perhaps you can move some of this early-exit code to a separate function.  A person reading the code should not be bogged down by details related to an uncommon use-case.

examples/classifier_compression/compress_classifier.py
```diff
@@ -409,14 +438,37 @@ def train(train_loader, model, criterion, optimizer, epoch,
...
+                         OrderedDict([
+                             ('Epoch', epoch),
+                             ('i', train_step),
+                             ('Objective Loss', losses['objective_loss'].mean),
+                             ('Prec@1_exit0', exit0err.value(1)),
+                             ('Prec@5_exit0', exit0err.value(5)),
+                             ('Prec@1_exit1', exit1err.value(1)),
+                             ('Prec@5_exit1', exit1err.value(5)),
+                             ('Prec@1_exitN', exitNerr.value(1)),
+                             ('Prec@5_exitN', exitNerr.value(5))]))
```
nzmora: Same comments as before:
1. This is not generic code (dependent on the number of exist.
2. There are 20 lines of code that won't be of interest to most readers.  Better push them to a separate function.

examples/classifier_compression/compress_classifier.py
```diff
@@ -460,31 +525,145 @@ def _validate(data_loader, model, criterion, loggers, print_freq, epoch=-1):
...
+            elif earlyexit: # imagenet
+                loss_exit0 = criterion(output[0], target_var)
+                loss_exit1 = criterion(output[0], target_var)
+                loss_exitN = criterion(output[0], target_var)
+
+                # measure accuracy and record loss
+                losses_exit0.add(loss_exit0.item())
+                losses_exit1.add(loss_exit1.item())
+                losses_exitN.add(loss_exitN.item())
+
```
nzmora: some of this code can be shortened and generalized if you used containers.  E.g.
```
for exit in num_exits:
      losses[exit].add(model_exit_loss[exit]).item())
```

models/cifar10/resnet_cifar_earlyexit.py
```diff
@@ -0,0 +1,188 @@
+#
```
nzmora: Do we need this entire file?  Can't you use the exiting model file and add a couple of 'if' statements?
Less code is less things to deal with.

models/imagenet/resnet_earlyexit.py
```diff
@@ -0,0 +1,243 @@
+import torch.nn as nn
```
nzmora: Same comment regarding the necessity of this file as for Resnet20 above.
Also, you added a new file and didn't document at the top why it is different from ResNet and why we need it.
Still, I prefer you try and use the existing model definition files.

examples/classifier_compression/compress_classifier.py
haim-barad: Both new parameters have default of "None" and there's no problem to move them to always be at the end of previous arguments.

examples/classifier_compression/compress_classifier.py
haim-barad: The parameter earlyexit isn't passed into that specific function. Should we rename lossweights to earlyexit_lossweights just to have "earlyexit" in the parameter? I thought it was too long, but perhaps it keeps the parameters together, so to speak.

examples/classifier_compression/compress_classifier.py
haim-barad: Agreed.

examples/classifier_compression/compress_classifier.py
guyjacob: Also, please keep the existing notations, no reason to have "Top1" in one case and "Prec@1" in the other

models/imagenet/resnet_earlyexit.py
haim-barad: As per Guy's comment below, this will be handled as a subclass.

models/cifar10/resnet_cifar_earlyexit.py
haim-barad: As per Guy's comment below, this will be handled as a subclass.

examples/classifier_compression/compress_classifier.py
```diff
@@ -212,6 +212,16 @@ def main():
...
+    if args.earlyexit_thresholds:
+        # the following is set in accordance with the example code for some models
+        if args.dataset == 'cifar10':
+            num_exits = 2
+        else:      # imagenet
+            num_exits = 3
+        loss_exits = [0] * num_exits
+        losses_exits = []
+        exiterrors = []
+
```
nzmora: This code is written for a specific network, so it is not generic for other cases.
Basically you can infer ```num_exits``` from ```len(earlyexit_lossweights)``` or len(earlyexit_thresholds), right?

examples/classifier_compression/compress_classifier.py
```diff
@@ -328,21 +338,20 @@ def main():
         # Train for one epoch
         train(train_loader, model, criterion, optimizer, epoch, compression_scheduler,
               loggers=[tflogger, pylogger], print_freq=args.print_freq, log_params_hist=args.log_params_histograms,
-              lossweights=args.lossweights, dataset=args.dataset)
+              earlyexit_lossweights=args.earlyexit_lossweights, num_exits=num_exits, exiterrors=exiterrors)
```
nzmora: To try and control the argument bloat, I've changed the interface of train(), validate(), test() to include ```args``

examples/classifier_compression/compress_classifier.py
```diff
@@ -689,6 +621,41 @@ def get_inference_var(tensor):
...
+    sum_lossweights = 0
+    for exitnum in range(num_exits-1):
+        loss += (earlyexit_lossweights[exitnum] * criterion(output[exitnum], target_var))
+        sum_lossweights += earlyexit_lossweights[exitnum]
+        exiterrors[exitnum].add(output[exitnum].data, target_var)
+    # handle final exit
+    loss += (1.0 - sum_lossweights) * criterion(output[num_exits-1], target_var)
+    exiterrors[num_exits-1].add(output[num_exits-1].data, target_var)
+    return loss
+
```
nzmora: On more careful reading of this code, I realize that this can be used in other use-case where we want to have weighted losses.

examples/classifier_compression/compress_classifier.py
```diff
@@ -526,144 +507,95 @@ def _validate(data_loader, model, criterion, loggers, print_freq, earlyexit, dat
...
-                    if exit_0:
-                        statsDict['Top1 exit0'] = exit0err.value(1)
-                        statsDict['Top5 exit0'] = exit0err.value(5)
-                    if exit_1:
-                        statsDict['Top1 exit1'] = exit1err.value(1)
-                        statsDict['Top5 exit1'] = exit1err.value(5)
-                    if exit_N:
-                        statsDict['Top1 exitN'] = exitNerr.value(1)
-                        statsDict['Top5 exitN'] = exitNerr.value(5)
+                    statsDict['LossAvg0'] = losses_exits[0].mean
```
nzmora: camel-case (statsDict) is frowned upon for variables in Python.  Better use ```stats_dict```

examples/classifier_compression/compress_classifier.py
```diff
@@ -526,144 +507,95 @@ def _validate(data_loader, model, criterion, loggers, print_freq, earlyexit, dat
...
+                    statsDict['LossAvgN'] = losses_exits[2].mean
+                    if exitstats[0]:
+                        statsDict['Top1 exit0'] = exiterrors[0].value(1)
+                        statsDict['Top5 exit0'] = exiterrors[0].value(5)
+                    if exitstats[1]:
+                        statsDict['Top1 exit1'] = exiterrors[1].value(1)
+                        statsDict['Top5 exit1'] = exiterrors[1].value(5)
+                    if exitstats[2]:
+                        statsDict['Top1 exitN'] = exiterrors[2].value(1)
+                        statsDict['Top5 exitN'] = exiterrors[2].value(5)
```
nzmora: Can't we replace these 3 ```if``` statements with:
```
for exit in range(len(exitstats)):
    stats_dict['Top1_exit{}'.format(exit)] = exit_errors[exit].value(1)
    stats_dict['Top5_exit{}'.format(exit)] = exit_errors[exit].value(5)
```

examples/classifier_compression/compress_classifier.py
```diff
@@ -526,144 +507,95 @@ def _validate(data_loader, model, criterion, loggers, print_freq, earlyexit, dat
...
-            validate_return[1] += exitNerr.value(5)
-            validate_return[2] += losses_exitN.mean
+        if exitstats[0]:
+            validate_return[0] += exiterrors[0].value(1)
+            validate_return[1] += exiterrors[0].value(5)
+            validate_return[2] += losses_exits[0].mean
+        if exitstats[1]:
+            validate_return[0] += exiterrors[1].value(1)
+            validate_return[1] += exiterrors[1].value(5)
+            validate_return[2] += losses_exits[1].mean
```
nzmora: Repetitive code like that can be scaled down.

examples/classifier_compression/compress_classifier.py
```diff
@@ -526,144 +507,95 @@ def _validate(data_loader, model, criterion, loggers, print_freq, earlyexit, dat
...
-        msglogger.info("Exit 0: %d", exit_0)
-        msglogger.info("Exit 1: %d", exit_1)
-        msglogger.info("Exit N: %d", exit_N)
-        msglogger.info("Percent Early Exit #0: %.3f", (exit_0*100.0) / (exit_0+exit_1+exit_N))
-        msglogger.info("Percent Early Exit #1: %.3f", (exit_1*100.0) / (exit_0+exit_1+exit_N))
+        msglogger.info("Exit 0: %d", exitstats[0])
+        msglogger.info("Exit 1: %d", exitstats[1])
+        msglogger.info("Exit N: %d", exitstats[2])
+        msglogger.info("Percent Early Exit #0: %.3f", (exitstats[0]*100.0) / (exitstats[0]+exitstats[1]+exitstats[2]))
+        msglogger.info("Percent Early Exit #1: %.3f", (exitstats[1]*100.0) / (exitstats[0]+exitstats[1]+exitstats[2]))
```
nzmora: Repetitive code like that can be scaled down.

examples/classifier_compression/compress_classifier.py
```diff
@@ -526,144 +507,95 @@ def _validate(data_loader, model, criterion, loggers, print_freq, earlyexit, dat
...
+            validate_return[1] += exiterrors[0].value(5)
+            validate_return[2] += losses_exits[0].mean
+        if exitstats[1]:
+            validate_return[0] += exiterrors[1].value(1)
+            validate_return[1] += exiterrors[1].value(5)
+            validate_return[2] += losses_exits[1].mean
+        if exitstats[2]:
+            validate_return[0] += exiterrors[2].value(1)
+            validate_return[1] += exiterrors[2].value(5)
+            validate_return[2] += losses_exits[2].mean
```
nzmora: Repetitive code like that can be scaled down.

examples/classifier_compression/compress_classifier.py
haim-barad: Yes - you could do that. OK - that's better than the assumptions I'm currently making (i.e. based on dataset only because that's is in accordance with the example).

examples/classifier_compression/compress_classifier.py
haim-barad: Good idea. The changes currently underway will use that. Note that I didn't previously address code bloat in _validate(), but I will do that now as well.

examples/classifier_compression/compress_classifier.py
haim-barad: Yes - as I mentioned previously, _validate() wasn't done until I got comments on train() - it now _validate() will be cleaned up.

PR #41, opened 2018-08-10
Knowledge distillation flow

guyjacob:
* Implemented as a Policy, with some caveats:
  - The teacher model is maintained in the policy, not returned to the user. So the script needs to call a dedicated `forward` function in the policy to facilitate forwarding through the teacher model.
  - At the moment it's not initialized via YAML. There's a "convenience" function to add command line arguments to an existing parser, but the teacher model and the policy need to be created "manually" by the script.
* Integrated in image classification sample

Still missing - documentation for the docs website


dalistarh:
Hi,

I believe a version of the distillation-for-quantization technique was concurrently developed by this paper:
https://openreview.net/forum?id=S1XolQbRW

(has a different quantization scheme though)

guyjacob:
@dalistarh Indeed I am familiar with your work. I'll make sure to add a reference to it in our documentation.
Please note that the distillation flow added here is not tied to any specific quantization method. It is enabled independently and can be used with any quantization method (or, for that matter, pruning / regularization method). 
Regardless, we appreciate your interest in this project, and we'll happy to hear any feedback you might have on the project in general, and this PR specifically.

dalistarh:
Hi Guy,

Thank you for this package, it's a really great idea coupled with great
execution. I think it will become the standard for research in this area.

I have a bunch of fine-grained praise and suggestions (I have been using it
a lot during the summer for my research), so I was wondering if there's a
contact email I should send them to, since they're not all relevant to this
PR.

Cheers,
Dan

guyjacob:
Hi Dan, thanks for your very kind words. I've reached out to you via e-mail.

Thanks,
Guy

PR #45, opened 2018-09-13
Add support for ResNext

haim-barad:
ResNext support added.


haim-barad:
Cancel PR for resnext or any future model as a static code insertion. Moving to a more general dynamic code modification for exit insertion.

PR #47, opened 2018-09-16
A small stand-alone utility to inspect the contents of checkpoint files

nzmora:


guyjacob:
Tried to run it I get `ImportError: No module named 'distiller'` when it tries to load the checkpoint. I think that since the checkpoints contain distiller objects, distiller should be explicitly imported.

Also - this is a generic utility, no? Maybe it should go in `apputils`? Or some dedicated scripts dir?

nzmora:
Checkpoints should not contain any Distiller class dependencies.  Are you trying some older checkpoint (there was a time when we had this dependency)?  Can you send it to me?

As far as its location: at the moment I don't want to open another directory.  ```apputils``` is a library, not an application directory, so I don't think it belongs there.

guyjacob:
If a checkpoint is saved using `apputils.save_checkpoint()`, then it could include a scheduler object, for example. That's what happens in `compress_classifier.py`. Or am I missing something?

nzmora:
```apputils.save_checkpoint()``` only stores dictionaries it gets from objects is serializes, such as CompressionScheduler instances.  In these dictionaries we only store Python and PyTorch types.
This is done carefully and intentionally in order to avoid creating a dependency on Distiller.  This is because these dependencies are very sensitive to the version of the code they are serializing (so sensitive that if you move file locations between Distiller versions, you won't be able to deserialize checkpoints).
In the past we had some dependencies so maybe you are using an old checkpoint file?  
Maybe we still have a bug somewhere and this is manifested in your use-case (and checkpoint file)?  -- send it over ;-)

guyjacob:
OK, dug a little deeper I think it's my bad - when I serialize the so-called `quantizer_metadata`, I store the quantizer's actual type (that is - the result of `type(self)`, not as a string. That's a Distiller type and apparently to deserialize a type object, the type it refers to has to exist. So - I wasn't following our practice :)

So this will only happen when quantization was used. I'll fix it - up to you if you want to wait with this PR.

Issue #49, opened 2018-09-19
Error in resume quantize checkpoint


chenys1995:
https://github.com/NervanaSystems/distiller/blob/e749ea6288431a53f839b621cc3e38facbf824de/distiller/quantization/range_linear.py#L165
I got an error message after resume symmetric linear quantized model
```
Traceback (most recent call last):
  File "compress_classifier.py", line 684, in <module>
    main()
  File "compress_classifier.py", line 244, in main
    model, chkpt_file=args.resume)
  File "/chenys/distiller/apputils/checkpoint.py", line 116, in load_checkpoint
    quantizer = qmd['type'](model, **qmd['params'])
TypeError: __init__() got an unexpected keyword argument 'bits_weights'
```
I modify the `__init__ ` argument and fix the problem
```python
def __init__(self, model, bits_activations=8, bits_weights=8, **kw):
     super(SymmetricLinearQuantizer, self).__init__(model, bits_activations=bits_activations,
                                                    bits_weights=bits_weights, 
                                                    train_with_fp_copy=False,
                                                    **kw)
```
I think the quantizer may be consistent in naming the parameters, or the quantizer_metadata will cause initialize error

guyjacob:
@chenys1995 this is a bug indeed - thanks for catching it! Your modification looks fine as well. Would you mind making a pull request with the fix? If that's not an option, let me know and I'll push the fix myself.

PR #50, opened 2018-09-20
Fix bug: Loading the SymmetricLinearQuantizer 

chenys1995:
Fix bug: error when loading SymmetricLinearQuantizer

### DIFF
--- a/distiller/quantization/range_linear.py
+++ b/distiller/quantization/range_linear.py
@@ -162,11 +162,15 @@ class SymmetricLinearQuantizer(Quantizer):
         model (torch.nn.Module): Model to be quantized
         bits_activations/parameters: Number of bits to be used when quantizing each tensor type
     """
-    def __init__(self, model, bits_activations=8, bits_weights=8, **kw):
+    def __init__(self, model, bits_activations=8, bits_parameters=8):
         super(SymmetricLinearQuantizer, self).__init__(model, bits_activations=bits_activations,
-                                                       bits_weights=bits_weights,
-                                                       train_with_fp_copy=False, **kw)
-
+                                                       bits_weights=bits_parameters,
+                                                       train_with_fp_copy=False)
+        
+        self.model.quantizer_metadata = {'type': type(self),
+                                         'params': {'bits_activations': bits_activations,
+                                                    'bits_parameters': bits_parameters}}
+        
         def replace_fn(module, name, qbits_map):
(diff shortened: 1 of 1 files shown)


distiller/quantization/range_linear.py
```diff
@@ -162,10 +162,10 @@ class SymmetricLinearQuantizer(Quantizer):
         model (torch.nn.Module): Model to be quantized
         bits_activations/parameters: Number of bits to be used when quantizing each tensor type
     """
-    def __init__(self, model, bits_activations=8, bits_parameters=8):
+    def __init__(self, model, bits_activations=8, bits_weights=8, **kw):
```
guyjacob: What's the reason for adding `**kw` here? Did you have a need to pass additional `Quantizer` keyword args that aren't exposed by `SymmetricLinearQuantizer`? I'm asking because I'd prefer to keep the arguments list explicit.

chenys1995: Since the __init__() of quantizer have write the metadata which is not in the SymmetricLinearQuantizer arguments list.
Maybe the better way is that writing its own metadata for each quantizer.

guyjacob: So far I chose to expose only the arguments that I thought were indeed optional for each specific method (that is - each subclass of `Quantizer`). If you think something is missing for this specific quantizer - let's discuss, and possibly add it explicitly. Otherwise, please remove the `**kw` argument for now.

distiller/quantization/range_linear.py
```diff
@@ -162,11 +162,15 @@ class SymmetricLinearQuantizer(Quantizer):
...
-    def __init__(self, model, bits_activations=8, bits_weights=8, **kw):
+    def __init__(self, model, bits_activations=8, bits_parameters=8):
         super(SymmetricLinearQuantizer, self).__init__(model, bits_activations=bits_activations,
-                                                       bits_weights=bits_weights,
-                                                       train_with_fp_copy=False, **kw)
-
+                                                       bits_weights=bits_parameters,
+                                                       train_with_fp_copy=False)
+        
+        self.model.quantizer_metadata = {'type': type(self),
```
guyjacob: Hmm, didn't think of this direction. It does make at least some sense that sub-classes will store their own metadata. I'm not sure having a different argument name is a good enough reason, but let's keep it this way for now. I need to think if maybe some further is in order.

guyjacob:
This one solves #49. Thanks again @chenys1995

Issue #70, opened 2018-11-04
lr_scheduler doesn't work when start training from checkpoint


stvreumi:
Hello:
I want to use `MultiStepMultiGammaLR` scheduler in my pruning lr_scheduler.
When I using the `compress_classifier.py` to pruning the `res_net20_cifar` from the begining and define the lr_scheduler in the `yaml` file, it works well.
But when I using the checkpoint to train and prune, the lr_scheduler defined in the `yaml` file doesn't work. The lr **doesn't decay** when the epoch achieve defined milestone.

I use the script below:
`python3 compress_classifier.py --arch resnet20_cifar  dataset/ -p=50 --lr=0.3 --epochs=150 -b 128 --compress=resnet20_cifar_ele_pruning.yaml -j=1 --vs 0 --deterministic --resume=logs/resnet20_cifar_baseline/checkpoint.pth.tar`

Below is my `yaml` setting
```
version: 1
pruners:
  low_pruner:
    class: AutomatedGradualPruner
    initial_sparsity : 0.05
    final_sparsity: 0.60
    weights: [module.layer1.2.conv1.weight,  module.layer1.2.conv1.weight,
              module.layer1.0.conv1.weight,  module.layer1.0.conv2.weight,
              module.layer1.1.conv1.weight,  module.layer1.1.conv2.weight]

  mid_pruner:
    class:  AutomatedGradualPruner
    initial_sparsity : 0.05
    final_sparsity: 0.67
    weights: [module.layer2.2.conv1.weight,  module.layer2.2.conv2.weight,
              module.layer2.0.conv2.weight,  module.layer2.0.downsample.1.weight,
              module.layer2.0.conv1.weight,  module.layer2.0.downsample.0.weight,
              module.layer2.1.conv1.weight,  module.layer2.1.conv2.weight]

  high_pruner:
    class:  AutomatedGradualPruner
    initial_sparsity : 0.05
    final_sparsity: 0.76
    weights: [module.layer3.0.conv1.weight,  module.layer3.1.conv1.weight,
              module.layer3.1.conv2.weight,  module.layer3.0.conv2.weight,
              module.layer3.0.downsample.0.weight, module.layer3.0.downsample.1.weight,
              module.fc.weight]
lr_schedulers:
  training_lr:
    class: MultiStepMultiGammaLR
    milestones: [300, 302, 400]
    gammas: [0.1, 0.1, 0.5]

policies:
    - pruner:
        instance_name: low_pruner
[...]

PR #72, opened 2018-11-05
explicitly adjust lr according to current epoch

hunterkun:


nzmora:
Nice and simple - I like this solution to issue #70.
I'll do some testing before merging.
Thanks!
Neta

netaz:
@guyjacob What do you think?  This is not perfect, but it's better than the current situation.

guyjacob:
TL;DR - I think we can go with this solution for now, even though it doesn't a specific scenario (details below). The proper solution we should implement is saving and loading the optimizer state.

It seems to me that the only case where this solution will not work is if the LR scheduler updates the LR "progressively". That is, the calculation is based on the current LR and not solely on the initial LR. In that case, if we don't restore the optimizer state as well (which contains the last updated LR), we would get unexpected results.
I checked, and as far as I can see the only scheduler that does such "progressive" updates is [`ReduceLROnPlateau`](https://github.com/pytorch/pytorch/blob/master/torch/optim/lr_scheduler.py#L254) (this includes the schedulers in PyTorch and in Distiller).
So I think we can live with this corner case for now, but as I wrote above, eventually we should implement proper save/load of optimizer state.

guyjacob:
Closing in favor of #74

PR #74, opened 2018-11-08
resume optimizer 

hunterkun:
This implementation resume optimizer but remember the lr scheduler adjust learning rate according to current epoch and initial lr rather than lr of your last training time.

### DIFF
--- a/distiller/config.py
+++ b/distiller/config.py
@@ -49,7 +49,7 @@ msglogger = logging.getLogger()
 app_cfg_logger = logging.getLogger("app_cfg")
 
 
-def dict_config(model, optimizer, sched_dict):
+def dict_config(model, optimizer, start_epoch, sched_dict):
     app_cfg_logger.debug('Schedule contents:\n' + json.dumps(sched_dict, indent=2))
 
     schedule = distiller.CompressionScheduler(model)
@@ -109,7 +109,7 @@ def dict_config(model, optimizer, sched_dict):
             add_policy_to_scheduler(policy, policy_def, schedule)
 
         # Any changes to the optmizer caused by a quantizer have occured by now, so safe to create LR schedulers
-        lr_schedulers = __factory('lr_schedulers', model, sched_dict, optimizer=optimizer)
+        lr_schedulers = __factory('lr_schedulers', model, sched_dict, optimizer=optimizer, last_epoch=start_epoch-1)
         for policy_def in lr_policies:
             instance_name, args = __policy_params(policy_def, 'lr_scheduler')
             assert instance_name in lr_schedulers, "LR-scheduler {} was not defined in the list of lr-schedulers".format(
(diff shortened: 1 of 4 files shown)


distiller/quantization/quantizer.py
```diff
@@ -176,6 +176,10 @@ def prepare_model(self):
         if self.optimizer:
             optimizer_type = type(self.optimizer)
             new_optimizer = optimizer_type(self._get_updated_optimizer_params_groups(), **self.optimizer.defaults)
+            # we need explicitly add the initial_lr attribute to new_optimizer
```
guyjacob: Good catch, wouldn't have thought about this one. I think to be on the safe side we should also set `'lr'`, not just `'initial_lr'`. Just in case some LR scheduler would use the current LR in addition to the initial one.

guyjacob:
Following the latest [comments](https://github.com/NervanaSystems/distiller/issues/70#issuecomment-436235921) in #70, we should make a distinction within `compress_classifier.py` between 2 cases:
1. We want to resume a previous training session. That is - we want to load the weights **AND** the hyper-parameters that are stored in the optimizer state.
2. We want to fine-tune an pre-trained model. That is - we only want to load the weights, and we're defining new hyper-parameters with a "fresh" optimizer.

Note that option (2) is exactly what happens today when `--resume` is passed to `compress_classifier.py`. So while the naming may be misleading, in order to not break existing code / scripts, let's keep it that way for now.

In order to support option (1) as well, **can you please add another command line option** - call it `--resume_w_state`. Have it with the same parameters as the existing resume option:

https://github.com/NervanaSystems/distiller/blob/248bca473ec1c46d18864ab6e57e3ed6d709e20d/examples/classifier_compression/compress_classifier.py#L115-L116

So  to do (1), the user will need to use `resume_w_state <chkpt_path>`. For (2) - `resume <chkpt_path>` as before. Then modify the code that handles resuming accordingly (line 307).

Please let me know if you have any questions, or if you prefer that we make this change from our side. Thanks!

distiller/config.py
```diff
@@ -137,13 +137,13 @@ def add_policy_to_scheduler(policy, policy_def, schedule):
                             frequency=policy_def['frequency'])
 
 
-def file_config(model, optimizer, filename):
+def file_config(model, optimizer, filename, start_epoch):
```
guyjacob: This should be optional as well
```suggestion
def file_config(model, optimizer, filename, start_epoch=0):
```

distiller/quantization/quantizer.py
```diff
@@ -176,6 +176,11 @@ def prepare_model(self):
         if self.optimizer:
             optimizer_type = type(self.optimizer)
             new_optimizer = optimizer_type(self._get_updated_optimizer_params_groups(), **self.optimizer.defaults)
+            # we need explicitly add the initial_lr and lr attribute
+            for group, new_group in zip(self.optimizer.param_groups, new_optimizer.param_groups):
+                if 'initial_lr' in group.keys():
+                    new_group['initial_lr'] = group['initial_lr']
+                new_group['lr]=group['lr]
```
guyjacob: Missing closing quotes
```suggestion
                new_group['lr'] = group['lr']
```

guyjacob:
Resolved in #182

PR #84, opened 2018-11-24
Quantization Features

guyjacob:
* Asymmetric post-training quantization (only symmetric supported so until now)
* Quantization aware training for range-based (min-max) symmetric and asymmetric quantization
* Per-channel quantization support in all of the above scenarios


guyjacob:
Oops - screwed up here with with rebasing on the master. Will open new PR instead.

PR #88, opened 2018-11-29
Fix assign_layer_fq_names

barrh:
Add support to models that contain named empty layers.
This solution suppresses such events, and the layer is ignored.

How to reproduce:
See the Resnet variants found on: https://github.com/Cadene/pretrained-models.pytorch
Specifically, the last FC layer

### DIFF
--- a/distiller/utils.py
+++ b/distiller/utils.py
@@ -72,8 +72,12 @@ def assign_layer_fq_names(container, name=None):
     """
     is_leaf = True
     for key, module in container._modules.items():
-        is_leaf = False
-        assign_layer_fq_names(module, ".".join([name, key]) if name is not None else key)
+        try:
+            assign_layer_fq_names(module, ".".join([name, key]) if name is not None else key)
+            is_leaf = False
+        except AttributeError:
+            if module is not None:
+                raise
     if is_leaf:
         container.distiller_name = name
 


barrh:
yeah, I'm sorry about that. I probably should have started with explaining the issue.
I have imported pretrained resnext and resnet models via https://github.com/Cadene/pretrained-models.pytorch (instead of Distiller's own models package), and executed inference on CPU.
I used distiller/jupyter/save_intermediate_feature_maps.ipynb as a reference, and I had modified it.

[save_intermediate_feature_maps.ipynb](https://github.com/NervanaSystems/distiller/files/2631067/save_intermediate_feature_maps.txt)
The file attached above is IPYNB (jupyter notebook) file, taken from the Distiller repo, and modified to import resnet50 via https://github.com/Cadene/pretrained-models.pytorch instead of through Distiller own models package.
Should you choose to execute it, it will fail while executing assign_layer_fq_names which was invoked with arguments: (None, 'FC').

nzmora:
Hi Bar,
I looked into this and the issue exists because the ResNet model from [Cadene's repository]( https://github.com/Cadene/pretrained-models.pytorch) attaches a ```NoneType``` to the ```fc``` layer (and adds a ```torch.nn.modules.linear.Linear``` named ```last_linear```).  
This is quite unusual, but since Cadene's repository]( https://github.com/Cadene/pretrained-models.pytorch) is gaining popularity, I think it makes sense to merge this fix.

Thanks!
Neta

PR #91, opened 2018-11-30
Support low precision weight(<=8 bits) saving

chenys1995:
I think this is helpful for demonstrate low precision quantization is useful for reducing the model size
although Pytorch still load into FP32 weights


SystemErrorWang:
@nzmora  i would also like to know how to save the weight file in 8bit, so it could be 4x smaller than a normal one saved in float32. i tried quantization-aware training and post-training quantization, but the saved models are still as big as norm.

nzmora:
Hi @SystemErrorWang,

Currently this is not supported.  See more details [here](https://github.com/NervanaSystems/distiller/issues/159#issuecomment-465079046).

Cheers
Neta

PR #92, opened 2018-11-30
Bug fixed for evaluating thinning and linear quantized model and Low-precisions checkpoint saving

chenys1995:
Since the low precision quantization does not save the weight to correspond data type.
It may be helpful to check the reduction of model size directly. 

SymmetricLinearQuantizer override the FP32 weight to int value
But it can be saved by int8 to reduce storage space
and will automatically load to FP32 by load_state_dict 

A simple usage example 
```python3 compress_classifier.py -a resnet20_cifar ./cifar-10 --qe -e --lowp ```
can easily compare reduction of size with 
```python3 compress_classifier.py -a resnet20_cifar ./cifar-10 --qe -e ```
by check the disk size

Another issue is that doing quantize for thinning model in evaluate_model(--quantize):
Because compression_scheduler is not saved by apputil.save_checkpoint,
It can not apply thinning recipe when loading checkpoint

### DIFF
--- a/apputils/checkpoint.py
+++ b/apputils/checkpoint.py
@@ -66,6 +66,17 @@ def save_checkpoint(epoch, arch, model, optimizer=None, scheduler=None,
         checkpoint['thinning_recipes'] = model.thinning_recipes
     if hasattr(model, 'quantizer_metadata'):
         checkpoint['quantizer_metadata'] = model.quantizer_metadata
+        b_wts = 'bits_weights' if 'bits_weights' in model.quantizer_metadata['params'] else 'bits_parameters'
+        if  model.quantizer_metadata['params'][b_wts] <= 8:
+            msglogger.info("Storing low precision state_dict")
+            q_dict = {}
+            for k, v in model.state_dict().items():
+                q_dict[k] = v.clone().detach()
+                q_dict[k].type(torch.int8)
+                verify = q_dict[k].clone().detach()
+                verify.type(torch.float32)
+                assert verify.equal(v)
+            checkpoint['state_dict'] = q_dict
 
     torch.save(checkpoint, fullpath)
     if is_best:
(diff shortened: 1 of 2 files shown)


guyjacob:
@chenys1995 - My apologies, I haven't had time to look at this yet. I'll try to get to it in the next day or two. Thank you for contributing!

chenys1995:
@guyjacob It's fine.
It give me more time to address the problems inside the previous code.
As I am a graduate TA, I just try to solve problems which student usually encounter

guyjacob:
Hi, apologies again for the delay.

I agree that at the moment we don't have a clear way of showing the memory savings from quantization. However, at the moment I'd prefer to wait a bit with merging a fix like you propose, in favor of more proper support for quantization that should come from PyTorch and ONNX (hopefully sooner than later). That's for 8-bit quantization.
In addition, for quantization below 8-bit I wouldn't go with a solution like you propose, as using int8 for storage might be "closer" than FP32, but still inaccurate. For those cases I think a better way to "communicate" the possible savings will be via a model summary. We could use the existing `model_performance_summary` (see [here](https://github.com/NervanaSystems/distiller/blob/master/distiller/model_summaries.py#L197)). That's invoked by passing `--summary compute` in the command line. If you wish to give something like that a go, that would be great.

Regarding the fix passing the scheduler to be saved in the quantized checkpoint - that's a valid fix. The only thing I'd ask is that you add a comment explaining why we're passing the scheduler to the evaluate method.

So - if you don't mind separating the scheduler fix (perhaps create a new PR), I'd be happy to merge it.

chenys1995:
The one of original purposes is want to demonstrate some benefit for quantization even pytorch or onnx isn't officially support low precision operator now.  
The other is that reduce disk storage, That is the reason I using the argument to control it
image
To purely compare the model size(state_dict), Indeed it is better for output two pth file in one directory via model summary

PR #95, opened 2018-12-03
"Range-Based" Linear Quantization Features

guyjacob:
- Asymmetric post-training quantization (only symmetric supported so until now)
- Quantization aware training for range-based (min-max) symmetric and asymmetric quantization
- Per-channel quantization support in both training and post-training


distiller/quantization/q_utils.py
```diff
@@ -17,15 +17,30 @@
...
-def asymmetric_linear_quantization_scale_factor(num_bits, saturation_min, saturation_max):
+def asymmetric_linear_quantization_params(num_bits, saturation_min, saturation_max,
+                                          integral_zero_point=True, signed=False):
     n = 2 ** num_bits - 1
-    return n / (saturation_max - saturation_min)
+    scale = n / (saturation_max - saturation_min)
+    zero_point = scale * saturation_min
+    if integral_zero_point:
+        if isinstance(zero_point, torch.Tensor):
+            zero_point = zero_point.round()
```
nzmora: It would be interesting to test stochastic rounding

distiller/quantization/q_utils.py
```diff
@@ -17,15 +17,30 @@
...
+        else:
+            zero_point = float(round(zero_point))
+    if signed:
+        zero_point += 2 ** (num_bits - 1)
```
nzmora: It's probably clearer if you wrote:
```
zero_point += n
```
BTW, why is this required for asymmetric quantization?

distiller/quantization/q_utils.py
nzmora: So zero is not an integral - doesn't this create bias when zero padding (like gemmlowp docs say)?

guyjacob: Specifically for the zero point? Didn't think of that. Not sure how much of a difference it would make. TBH we don't have stochastic rounding at all in Distiller. There are arguments for and against it. But in the spirit of enabling experimentation we should probably do it. But not in this PR.

distiller/quantization/q_utils.py
guyjacob: I'll change. It's not required, it's just an option to have the quantized range with a signed integer instead of unsigned. Might be that some HW implementation / ISA needs signed data types, for example. Unsigned is the default.

distiller/quantization/q_utils.py
guyjacob: Indeed, integral zero point is the best practice, and is the default. Again, it's just an option to turn it off if someone wants to experiment and see the effects. If you're arguing that the optional parameter should be on the negative, meaning we should have a `non_integral_zero_point` parameter instead - I might be inclined to agree.

distiller/quantization/q_utils.py
guyjacob: Correction - actually, `n = 2 ** num_bits - 1`, so what we're adding to the zero point isn't `n`.

PR #96, opened 2018-12-06
Bug fixed for earlyexit

Gxllii:
I found that the script reports errors when I train earlyexit networks due to the wrong number of exits.
`python3 compress_classifier.py --arch resnet20_cifar_earlyexit ../../../data.cifar10 -p 30 -j=1 --lr=0.01 --earlyexit_lossweights 0.1 0.3 --earlyexit_thresholds 1.2 0.9`

And I updated the examples of earlyexit arguments which were not consistent with descriptions.


nzmora:
@Gxllii,

Why did you close the PR? :-) 
We're looking at it.
Cheers,
Neta

Gxllii:
@nzmora 
Oh, maybe the exits number is right.
We should use only 1 argument when the number of exits is 2.
I'm checking it.

Gxllii:
OK, thanks. 
The documentation will be helpful.

PR #98, opened 2018-12-08
delete the unnecessary colon

103yiran:


guyjacob:
@103yiran Thanks for spotting it.

PR #99, opened 2018-12-09
Fix bug: --qe do not save compression scheduler

chenys1995:
Bug: quantize a model which have thining_recipe  => loading checkpoint fail
The compression scheduler did not pass and save into checkpoint when --qe -e is set
It cause loading the checkpoint file failed since the thinning_recipe should combine with compression_ scheduler


guyjacob:
Merged, thanks @chenys1995

PR #104, opened 2018-12-11
update requirements.txt

Gxllii:
pytorch version should be >= 0.4.1
`earlyexit_validate_criterion = nn.CrossEntropyLoss(reduction='none').cuda()
TypeError: init() got an unexpected keyword argument 'reduction'`
keyword "reduction" cannot be used in pytorch 0.4.0 .


nzmora:
Hi @Gxllii ,

There was a misunderstanding - sorry about that!
The Early Exit code inserted a dependency on PyTorch 0.4.1, but this version is not supported by the rest of Distiller yet (we would need to do proper testing of all flows before changing requirements.txt).
This should be fixed very soon.

Cheers,
Neta

Gxllii:
OK, thanks.

PR #114, opened 2018-12-31
Fix best top1

barrh:
This patch tries to solve an issue I experienced when pruning pretrained models using distiller.
'best epochs' does not distinguish between top1 measurements taken before the model was pruned, or after, which results in bias toward the dense model measurements, and undermines the purpose of the study.
This patch implements the behavior I would expect to see.

Please note, I didn't test that patch thoroughly.

### DIFF
--- a/examples/classifier_compression/compress_classifier.py
+++ b/examples/classifier_compression/compress_classifier.py
@@ -383,8 +383,7 @@ def main():
             teacher, _, _ = apputils.load_checkpoint(teacher, chkpt_file=args.kd_resume)
         dlw = distiller.DistillationLossWeights(args.kd_distill_wt, args.kd_student_wt, args.kd_teacher_wt)
         args.kd_policy = distiller.KnowledgeDistillationPolicy(model, teacher, args.kd_temp, dlw)
-        compression_scheduler.add_policy(args.kd_policy, starting_epoch=args.kd_start_epoch, ending_epoch=args.epochs,
-                                         frequency=1)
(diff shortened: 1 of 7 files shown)


examples/classifier_compression/compress_classifier.py
```diff
@@ -426,18 +427,21 @@ def main():
...
-            # Keep best_epochs sorted such that best_epochs[0] is the lowest top1 in the best_epochs list
-            best_epochs = sorted(best_epochs, key=lambda score: score.top1)
-        for score in reversed(best_epochs):
+        # Update the list of top scores achieved since all policies have concluded
+        if top1 > best_epochs[-1].top1:
+            best_epochs[-1].epoch = epoch
+            best_epochs[-1].top1 = top1
+            best_epochs[-1].top5 = top5
+            # Keep best_epochs sorted from best to worst
+            best_epochs = sorted(best_epochs, key=operator.attrgetter('top1', 'top5'), reverse=True)
```
nzmora: Please add a comment saying: "sort by top1, and use top5 as tie-breaker"

barrh:
While my solution might be too complex, I don't see how measuring sparsity alone would serve the purpose to the same extent. In my patch, I also take care of quantization.

I made some small changes to the first commit, but 2nd commit should be essentially untouched.

It's another discussion, but is top1's error rate more interesting than top5's?

barrh:
1. The last push has not been tested yet.
2. Besides the core fix, I've had a bunch of small changes in the 2nd commit.
3. I'd noticed thinning policy inherits from base policy class, but never initiates it... Is it a bug?

barrh:
I've opened another pull request.

PR #117, opened 2019-01-03
CPU inference

galnov:
Enabling partial CPU-only support, per a new input argument and available hardware.


models/__init__.py
```diff
@@ -81,6 +81,7 @@ def create_model(pretrained, dataset, arch, parallel=True, device_ids=None):
     elif parallel:
         model = torch.nn.DataParallel(model, device_ids=device_ids)
 
-    if torch.cuda.is_available():
+    if device_ids != -1:
```
nzmora: This 'if' statement is a bit nasty because it is True when device_ids is None, which is the default.
There are all kinds of places, mainly Notebooks but not only, where create_model is called and device_ids is set to None.  

I think we need a better way to decide where to place the model.

galnov: In order not to introduce a change that breaks the API, let's use the following device_ids logic:
None - use GPU if available, otherwise CPU
-1 - use CPU
>=0 - use GPU device IDs

PR #118, opened 2019-01-06
Irregular block pruning

nzmora:
Block pruning: support specifying the block shape from the YAML file

Block pruning refers to pruning 4-D structures of a a specific shape.
This is similar to filter/channel pruning but allows for non-regular
shapes that accelerate inference on a specific hardware platform.

If we want to introduce sparsity in order to reduce the compute load
of a certain layer, we need to understand how the HW and SW perform
the layer's operation and what vector shape is used.  Then we can
induce sparsity to math the vector shape.

For example, Intel AVX-512 are SIMD instructions that apply the same
instruction (Single Instruction) on a vector of inputs (Multiple
Data).  The following single instruction performs an element-wise
multiplication of two 16 32-bit elements:

     __m256i result = __mm256_mul_epi32(vec_a, vec_b);

If either vec_a or vec_b are partially sparse, we still need to perform
the multiplication operation and the sparsity does not help reduce the
compute latency.  However, if either vec_a or vec_b contain only zeros
then we can eliminate entirely the instruction.  In this case, we
would like to have sparsity of 16-elements.

It is a bit more complicated because we also need to understand how the
software maps layer operations to hardware.  For example, a 3x3
convolution can be computed as a direct-convolution, as a matrix multiply
operation, or as a Winograd matrix operation (to name a few ways of
computation).  These low-level operations are then mapped to SIMD
instructions.

Finally, the low-level SW needs to support a block-sparse storage-format
for the weight tensors (see for example:
http://www.netlib.org/linalg/html_templates/node90.html)


nzmora:
We ended up merging @barrh's 'block_pruning' branch, which was based on this branch.  Therefore we can close.

PR #121, opened 2019-01-09
Hotfix best checkpoint

barrh:
Alternative approach to fix the best checkpoint pruning issue.
This does not fix the issue with quantizers.
[Not tested yet]


barrh:
This is tested with master branch and works as expected.

barrh:
abandoned due to lack of reviews. See PR #132 , which is a better solution imo.

PR #127, opened 2019-01-15
Dorefa1 bit

JoyFYan:
I try to add `BinaryQuantizeSTE` to support handling for 1-bit weights.

### DIFF
--- a/distiller/quantization/clipped_linear.py
+++ b/distiller/quantization/clipped_linear.py
@@ -127,7 +127,7 @@ class WRPNQuantizer(Quantizer):
 
 def dorefa_quantize_param(param_fp, param_meta):
     if param_meta.num_bits == 1:
-        out = BinaryQuantizeSTE.apply(param_fp)
+        out = DorefaParamsBinarizationSTE.apply(param_fp)
     else:
         scale, zero_point = asymmetric_linear_quantization_params(param_meta.num_bits, 0, 1, signed=False)
         out = param_fp.tanh()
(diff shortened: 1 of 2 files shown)


distiller/quantization/clipped_linear.py
```diff
@@ -125,6 +125,16 @@ def relu_replace_fn(module, name, qbits_map):
 
         self.replacement_factory[nn.ReLU] = relu_replace_fn
 
+def dorefa_quantize_param(param_fp, param_meta):
```
guyjacob: This is fine, but you should also delete the original function just below.

distiller/quantization/q_utils.py
```diff
@@ -100,6 +100,18 @@ def get_quantized_range(num_bits, signed=True):
         return -n, n - 1
     return 0, 2 ** num_bits - 1
 
+class BinaryQuantizeSTE(torch.autograd.Function):
```
guyjacob: What you implemented here is specifically **weight binarization in DoReFa.** That's perfectly fine, but the class name implies that this is a generic implementation of binarization. So please rename it to reflect it's specific purpose, e.g. `DorefaParamsBinarizationSTE`.
Also, for the same reason, please move it to `clipped_linear.py` next to the `dorefa_quantize_param` function. It's not re-usable enough at the moment to be in `q_utils.py`.

distiller/quantization/q_utils.py
```diff
@@ -100,6 +100,18 @@ def get_quantized_range(num_bits, signed=True):
...
+    def forward(ctx, input, inplace=False):
+        if inplace:
+            ctx.mark_dirty(input)
+        E = input.abs().mean()
+        output = input.sign() * E
+        return output
+    
+    @staticmethod
+    def backward(ctx, grad_output):
+        return grad_output, None, None, None, None
```
guyjacob: The number of values returned from `backward` needs to match the number of input arguments to `forward` (see PyTorch [docs](https://pytorch.org/docs/stable/autograd.html#torch.autograd.Function.backward)). So it should be:
```suggestion
        return grad_output, None
```

guyjacob:
Looks good, merged. Thanks again!

PR #132, opened 2019-01-23
Fix best top1

barrh:
I re-open this pull request since the hotfix I've suggested has yet to be merged and I believe this to be permanent all-around solution to the issue at hand, and therefore better than the hotfix.
I also added minor bug fixes in that branch (commit: Refactor policies addition).


barrh:
abandoned.

PR #136, opened 2019-01-24
Post-train quantization based on stats + additional modules quantized

guyjacob:
This is a big PR, will be easiest to review commit-by-commit. Not exactly best-practice, but as it is I prefer this to separate PRs.

Summary of changes:
(1) Post-train quantization based on pre-collected statistics
(2) Quantized concat, element-wise addition / multiplication and embeddings
(3) Move post-train quantization command line args out of sample code
(4) Configure post-train quantization from YAML for more fine-grained control


distiller/data_loggers/collector.py
```diff
@@ -272,16 +279,153 @@ def to_xlsx(self, fname):
...
+        records['shape'] = ''
+        return records
+
+    def __init__(self):
+        # We don't know the number of inputs at this stage so we defer records creation to the actual callback
+        self.inputs = []
+        self.output = self.create_records_dict()
+
+
+class QuantCalibrationStatsCollector(ActivationStatsCollector):
```
nzmora: How is this class different from RecordsActivationStatsCollector?  Can you give a few points?

examples/classifier_compression/compress_classifier.py
```diff
@@ -190,22 +187,30 @@ def main():
     if args.summary:
         return summarize_model(model, args.dataset, which_summary=args.summary)
 
+    activations_collectors = create_activation_stats_collectors(model, *args.activation_stats)
+
+    if args.qe_calibration:
+        msglogger.info('Quantization calibration stats collection enabled: Setting constant seeds and converting '
+                       'model to serialized execution')
+        distiller.set_deterministic()
```
nzmora: why does it need to be deterministic?

examples/classifier_compression/compress_classifier.py
```diff
@@ -190,22 +187,30 @@ def main():
...
+        activations_collectors.update(create_quantization_stats_collector(model))
+
     # Load the datasets: the dataset to load is inferred from the model name passed
     # in args.arch.  The default dataset is ImageNet, but if args.arch contains the
     # substring "_cifar", then cifar10 is used.
     train_loader, val_loader, test_loader, _ = apputils.load_data(
         args.dataset, os.path.expanduser(args.data), args.batch_size,
-        args.workers, args.validation_size, args.deterministic)
+        args.workers, args.validation_size, args.deterministic,
+        shuffle_test=args.qe_calibration is not None)
```
nzmora: I'm not sure I like this here... so I might move it in the near future.
The reason is that the logic whether to shuffle or not might be more complicated (different flows -each making a different decision whether to shuffle or not).  On line 192 you have ```if args.qe_calibration:``` so you can move the shuffling decision there (i.e. group all the logic based on this ```if``` statement in one place).

examples/classifier_compression/compress_classifier.py
```diff
@@ -190,22 +187,30 @@ def main():
...
                    len(train_loader.sampler), len(val_loader.sampler), len(test_loader.sampler))
 
-    activations_collectors = create_activation_stats_collectors(model, *args.activation_stats)
-
     if args.sensitivity is not None:
         sensitivities = np.arange(args.sensitivity_range[0], args.sensitivity_range[1], args.sensitivity_range[2])
         return sensitivity_analysis(model, criterion, test_loader, pylogger, args, sensitivities)
 
-    if args.evaluate:
+    if args.evaluate or args.qe_calibration:
```
nzmora: I wonder what's clearer: the code as you wrote it or:
```
# add to the 'if' clause in line 192
if args.qe_calibration:
   args.evaluate = True
```

examples/classifier_compression/compress_classifier.py
```diff
@@ -433,13 +438,18 @@ def _validate(data_loader, model, criterion, loggers, args, epoch=-1):
     if args.display_confusion:
         confusion = tnt.ConfusionMeter(args.num_classes)
     total_steps = total_samples / batch_size
+    if args.qe_calibration:
```
nzmora: This is another example of what I meant earlier: we have a feature that allows us to use part of the dataset while training/validating/testing (I hope you added support for all these cases, because I need them).
The code here should set the total number of steps to the total configured which ranges between 100% to some_value.  The value of *some_value* is set by different logic.
Here's how I coded it:
```
total_samples = int(len(train_loader.sampler) * args.training_epoch_duration)
batch_size = train_loader.batch_size
steps_per_epoch = math.ceil(total_samples / batch_size)
```
```args.training_epoch_duration``` controls how much of the DS we use.  If you want to set it to a specific value do it somewhere else (e.g. line 192), but not here.  This will allow different logical paths to set ```args.training_epoch_duration``` to different values.

File: distiller/utils.py [hunk 6]
```diff
@@ -520,3 +524,29 @@ def replace_data_parallel(container):
...
+
+
+def set_deterministic():
+    torch.manual_seed(0)
+    random.seed(0)
+    np.random.seed(0)
+    torch.backends.cudnn.deterministic = True
+
+
+def yaml_ordered_load(stream, Loader=yaml.Loader, object_pairs_hook=OrderedDict):
```
nzmora: Why did you move this function here?  I only see it used in one file (config.py).  
Besides, utils.py is getting really bloated and we need to think how we refactor it.

examples/classifier_compression/compress_classifier.py
```diff
@@ -592,9 +592,14 @@ def evaluate_model(model, criterion, test_loader, loggers, activations_collector
 
     if args.quantize_eval:
         model.cpu()
-        quantizer = quantization.PostTrainLinearQuantizer(model, args.qe_bits_acts, args.qe_bits_wts,
-                                                          args.qe_bits_accum, args.qe_mode, args.qe_clip_acts,
-                                                          args.qe_no_clip_layers, args.qe_per_channel)
+        if args.qe_config_file:
```
nzmora: maybe consider moving this code into a separate function.

distiller/modules/eltwise.py
```diff
@@ -0,0 +1,50 @@
...
+
+import torch.nn as nn
+
+
+class EltwiseAdd(nn.Module):
+    def __init__(self, inplace=False):
+        super(EltwiseAdd, self).__init__()
+        self.inplace = inplace
+
+    # def forward(self, t1, t2):
```
nzmora: remove this remark?

distiller/data_loggers/collector.py
```diff
@@ -231,15 +235,18 @@ def to_np(stats):
         batch_min_list = to_np(torch.min(act, dim=1)).tolist()
         batch_max_list = to_np(torch.max(act, dim=1)).tolist()
         batch_mean_list = to_np(torch.mean(act, dim=1)).tolist()
-        batch_std_list = to_np(torch.std(act, dim=1)).tolist()
+        if act.shape[1] == 1:
```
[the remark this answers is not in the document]
guyjacob: `std` on a tensor with a single element returns `nan`. So just returning zeros in that case (that is - if the activation itself has a size of 1).
Coming to think of it, the condition here is a bit too specific. Only fits shapes of (N, 1). Something like: `if act.shape[0] == act.numel()` is probably better. I'll change.

distiller/data_loggers/collector.py
```diff
@@ -90,12 +87,12 @@ def start(self):
...
     def start_module(self, module):
-        """Iteratively register to the forward-pass callback of all eligable modules.
+        """Iteratively register to the forward-pass callback of all eligible modules.
 
-        Eligable modules are currently filtered by their class type.
+        Eligible modules are currently filtered by their class type.
         """
         is_leaf_node = len(list(module.children())) == 0
-        if is_leaf_node and type(module) in self.classes:
+        if is_leaf_node and (not self.classes or type(module) in self.classes):
```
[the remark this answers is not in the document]
guyjacob: You mean something like:
`if is_leaf_node and ((not self.classes) or type(module) in self.classes):`
?

distiller/data_loggers/collector.py
guyjacob: It tracks the values per-module, not per-activation. Actually the main reason why per-activation records aren't good enough will be when I add standard deviation. Here we need the std over all the activations, across all batches. So per-activation std doesn't help in this case. For all the other stats (at least the ones I'm collecting now) you can just apply the function to the per-activation values and that would be fine (assuming the sizes don't change between batches).
There are a couple of other differences, which could probably taken care of within `RecordsActivationStatsCollector`:

- Collect for inputs in addition to output
- Different serialization (YAML vs XLS, single value per-tensor vs. per-activation)

(Also less memory overhead because it saves just a single value)

distiller/utils.py
guyjacob: I also use it in `range_linear.py` now to load activation stats:
https://github.com/NervanaSystems/distiller/blob/1f10b9e8e142cef018740d1d12df9be254916d99/distiller/quantization/range_linear.py#L602-L603
Regardless, I agree we should do something about `utils.py`.

examples/classifier_compression/compress_classifier.py
guyjacob: Agreed. Shall I add a command-line argument to shuffle the test set as well? Or taking care of it internally good enough for your use case? We have so many command-line args already...

distiller/config.py
```diff
@@ -143,11 +143,12 @@ def file_config(model, optimizer, filename, scheduler=None):
     with open(filename, 'r') as stream:
         msglogger.info('Reading compression schedule from: %s', filename)
         try:
-            sched_dict = yaml_ordered_load(stream)
+            sched_dict = distiller.utils.yaml_ordered_load(stream)
             return dict_config(model, optimizer, sched_dict, scheduler)
         except yaml.YAMLError as exc:
             print("\nFATAL parsing error while parsing the schedule configuration file %s" % filename)
             raise
+            raise
```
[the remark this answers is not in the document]
guyjacob: This is a weird merge artifact in GitHub, the code on the branch doesn't actually have the redundant statement. So I can't fix it now. If it sticks after the merge I'll remove it then.

distiller/data_loggers/collector.py
guyjacob: Added some comments in the code explaining the differences

examples/classifier_compression/compress_classifier.py
guyjacob: Kept as-is for now

examples/classifier_compression/compress_classifier.py
guyjacob: Fixed

examples/classifier_compression/compress_classifier.py
guyjacob: Fixed

examples/classifier_compression/compress_classifier.py
guyjacob: Will be taken care of by merging the changes from 'master' that take care of using part of the dataset.

examples/classifier_compression/compress_classifier.py
guyjacob: Fixed

distiller/modules/eltwise.py
guyjacob: Fixed

PR #137, opened 2019-01-24
Energy measurement

joaolcaas:
Trying to measure energy in quantized and not quantized model (using nvidia from NVLM). 
Tested and working. Just having some questions about the results.


nzmora:
I'll try to integrate this PR sometime soonish ;-)
Cheers
Neta

nzmora:
Hi @joaolcaas ,

I'm closing this for now because it has been sitting stale for some time.
I think that GPU instrumentation should be performed in a process parallel to the main training loop, so that we collect instrumentation samples during the entire training/evaluation process, not only when the CPU core is running specific part of the training-loop.  The training-loop may have start/stop event triggers, but should not sample, because that creates bias in our sampling.
If you want to revisit this code, please let me know and we can discuss.

Cheers
Neta

Issue #124, opened 2019-01-12
Support for PyTorch 1.0


nzmora:
Distiller should support PyTorch 1.0.

Related: https://github.com/NervanaSystems/distiller/issues/85#issuecomment-453733443

promach:
For those computer systems without support for pytorch version 0.40 , you could try this [online colab example](https://colab.research.google.com/drive/1JpTC-22qNIn1DXNaePwocG1snghuJVFr)

barrh:
Pytorch v1 introduces breaking changes, so code following this change will not be backward-compatible. Since this is the case, it's worth _considering_ to overlap this effort with other breaking changes as well:
1. requirements.txt makeover - some packages are very old, e.g. tensorflow==1.7.0
2. Require Python v3.6(/3.7) or later - this will require no immediate code changes, and will open to door to integrate new Python features.
3. API changes (e.g. validation-size=0), etc.

barrh:
Regarding Py3.6, as a user of Distiller, since the bottleneck (for my jobs) is GPUs, I probably won't notice any improvement. As a developer on the other hand, the standard dict is a lot better (no more OrderedDict()!), async is stable API, and there's new formatting. (see relevant doc - https://docs.python.org/3/whatsnew/3.6.html)
Same deal with v3.7, nice new features for developers (debug hook, data classes, dict implementation standardized), but user gets mostly headaches, because it's not as popular yet.

API changes - ADC perhaps?
Also, I feel **print-freq** should be modified or supplemented by prints-per-epoch. I can never tell if print-freq=X is good, too much, or too few, for a given batch-size and dataset.

nzmora:
Release 0.3 with support for PyTorch 1.0.1 is out!  
Cheers
Neta

PR #139, opened 2019-01-30
Fixes for pytorch==1.0.0

levzlotnik:


tests/test_infra.py
```diff
@@ -13,18 +13,19 @@
...
-
 import logging
 import os
 import sys
 import pytest
 module_path = os.path.abspath(os.path.join('..'))
 if module_path not in sys.path:
     sys.path.append(module_path)
-
+import distiller
```
nzmora: If we're already changing code, let's do it thusly:

```
try:
    import distiller
except ImportError:
    module_path = os.path.abspath(os.path.join('..'))
    if module_path not in sys.path:	
         sys.path.append(module_path)	    
    import distiller
```
Once we merge your code for installing the distiller package, ```import distiller``` will throw an ImportError only for those using the package in development-mode (the only available mode today).  The code should "tell" this story...

tests/test_infra.py
```diff
@@ -13,18 +13,19 @@
...
 from models import create_model
 from apputils import load_checkpoint
 
+
```
nzmora: The imports section should be separated from the code section by exactly 2 blank lines.

tests/test_summarygraph.py
```diff
@@ -61,24 +61,19 @@ def test_connectivity():
     assert g is not None
 
     op_names = [op['name'] for op in g.ops.values()]
-    assert 73 == len(op_names)
+    assert 80 == len(op_names)
```
nzmora: Can you explain why there are more operations now?
I understand it is related to PyTorch 1.0, but do you know what operations were added?
It's important for us to understand what has changed with PyTorch 1.0, because it will help us debug when issues arise.

levzlotnik: It's from ONNX... the graph description was changed

distiller/model_summaries.py
```diff
@@ -20,14 +20,16 @@
     - optimizer state
     - model details
 """
+import distiller
```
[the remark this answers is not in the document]
levzlotnik: I think that moving the functions to apputils/model_summaries.py would be beneficial, since both of the files pass the same functionality. Maybe even implementing these in terms of SummaryGraph?
What do you think?

nzmora:
@everyone Note that this PR is related to issue #124

guyjacob:
Completed in #144

PR #141, opened 2019-01-31
Refactor load_checkpoint

barrh:
I have noticed recent commit on load_checkpoint, and since I have already refactored it recently, I propose my changes.

The original purpose was to support loading converted-checkpoints that do not contain the checkpoint structure that Distiller expects. However, this pull request is stripped from the converted-checkpoints support, because I don't think it should be merged into master.

[I've never tested this commit]

### DIFF
--- a/tests/test_infra.py
+++ b/tests/test_infra.py
@@ -17,6 +17,9 @@
 import logging
 import os
 import sys
+import tempfile
+
+import torch
 import pytest
 module_path = os.path.abspath(os.path.join('..'))
 if module_path not in sys.path:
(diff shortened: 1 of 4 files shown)


nzmora:
Hi Bar,

It's hard to understand what you changed.  Look at the diff - the entire code look is red/green so it's hard to tell what you actually changed.
You also didn't test it... so why push it as a PR?
Thanks
Neta

tests/test_infra.py
```diff
@@ -34,6 +37,29 @@ def test_load():
...
+def test_load_state_dict():
+    # prepare lean checkpoint
+    state_dict_arrays = torch.load('../examples/ssl/checkpoints/checkpoint_trained_dense.pth.tar').get('state_dict')
+
+    with tempfile.NamedTemporaryFile() as tmpfile:
+        torch.save({'state_dict': state_dict_arrays}, tmpfile.name)
+        model = create_model(False, 'cifar10', 'resnet20_cifar')
+        model, compression_scheduler, start_epoch = load_checkpoint(model, tmpfile.name)
+
+    assert len(list(model.named_modules())) >= len([x for x in state_dict_arrays if x.endswith('weight')]) > 0
```
barrh: Neta, can you suggest better check here? (Line 49)

barrh:
Following recent update to my branch (barrh:load_cp), there seems to be conflicts in this PR.
This PR is already superseded by at least another PR.

PR #143, opened 2019-02-06
Relax repo requirements

barrh:
Add relaxation to requirements.txt .
The only strict requirements I left are: torch and pytest libraries.

I opted not to upgrade pytest because it introduces new deprecation warnings, that aren't printed with pytest v3.5.1. Looking into the warnings in my environment, it does not seem to be a bug in Distiller.

I ran pytest (with Py3.6) and it ran successfully.


nzmora:
Hi @barrh,

["In building your Python application and its dependencies for production, you want to make sure that your builds are predictable and deterministic. Therefore, always pin your dependencies."](https://nvie.com/posts/pin-your-packages/) -- this is what we're trying to practice.
We will upgrade the library versions with the next release.

Cheers
Neta

barrh:
If you want to achieve determinism, why not upload a docker image?

This approach does not generalize well. It only guarantees conflicts between different libraries that apply it asynchronously.
Bugs, on the other hand, are not necessarily deterministic.

PR #144, opened 2019-02-06
Release 0.3 (pytorch 1.0.0)

levzlotnik:
Refactoring code pre-release


examples/classifier_compression/compress_classifier.py
```diff
@@ -730,11 +726,11 @@ def save_collectors_data(collectors, directory):
 
 
 def check_pytorch_version():
-    if torch.__version__ < '0.4.0':
+    if torch.__version__ < '1.0.0':
```
guyjacob: We're going for 1.0.1, right?

examples/word_language_model/main.py
```diff
@@ -24,7 +24,7 @@
 if module_path not in sys.path:
```
guyjacob: Here also, `script_dir` and `module_path` aren't required for imports. They're not used for anything else, so please remove them.

jupyter/distiller_jupyter_helpers.ipynb
```diff
@@ -18,13 +18,13 @@
     "# Relative import of code from distiller, w/o installing the package\n",
     "import os\n",
     "import sys\n",
-    "module_path = os.path.abspath(os.path.join('..'))\n",
-    "if module_path not in sys.path:\n",
-    "    sys.path.append(module_path)\n",
+    "#module_path = os.path.abspath(os.path.join('..'))\n",
+    "#if module_path not in sys.path:\n",
+    "#    sys.path.append(module_path)\n",
```
guyjacob: Any reason to comment and not remove these lines?

requirements.txt
```diff
@@ -1,19 +1,19 @@
...
+torch==1.0.1
+numpy>=1.14.3
 torchvision==0.2.1
-scipy==1.1.0
+scipy>=1.1.0
 gitpython
 torchnet==0.0.4
-tensorflow==1.7.0
+tensorflow>=1.7.0
 pydot==1.2.4
```
guyjacob: Why aren't we trying to move all packages to their latest versions? @nzmora?

barrh: +1 for relaxing requirements **on master branch** (versus releases)
By the way, do we really need tensorflow package? won't tensorboard be enough?

examples/classifier_compression/compress_classifier.py
```diff
@@ -730,11 +726,11 @@ def save_collectors_data(collectors, directory):
 
 
 def check_pytorch_version():
-    if torch.__version__ < '0.4.0':
+    if torch.__version__ < '1.0.1':
```
barrh: This string comparison is bug-prone. Can you suggest another method to test the version?
e.g. '1.0.1.post10' < '1.0.1.post2' (current latest)
e.g. '10.0' < '2.0'

requirements.txt
```diff
@@ -1,19 +1,19 @@
-torch==0.4.0
-numpy==1.14.3
+torch==1.0.1
+numpy>=1.14.3
 torchvision==0.2.1
-scipy==1.1.0
+scipy>=1.1.0
 gitpython
```
nzmora: Let's set the version explicitly as '2.1.11'

requirements.txt
nzmora: > Why aren't we trying to move all packages to their latest versions? @nzmora?
We should - I thought this was the decision.  But I say: let's release the code, and then update these files.  Otherwise, we need to test everything with the new packages.

tests/test_summarygraph.py
```diff
@@ -19,13 +19,19 @@
...
 import sys
 import pytest
-module_path = os.path.abspath(os.path.join('..'))
-if module_path not in sys.path:
-    sys.path.append(module_path)
-import distiller
-from models import ALL_MODEL_NAMES, create_model
-from apputils import *
-from distiller import normalize_module_name, denormalize_module_name
+try:
```
nzmora: Why do we still need this try/except block once we use "pip install -e"?
@guyjacob - thoughts?

examples/classifier_compression/compress_classifier.py
barrh: @nzmora Without getting into how string comparison implemented, simply put, comparing strings isn't equivalent to comparing versions. It's only by mere chance that it worked until now. see counter examples above.

tests/test_summarygraph.py
guyjacob: We don't. Guess I missed it when I reviewed the changes.

jupyter/truncated_svd.ipynb
```diff
@@ -216,8 +214,8 @@
     "import time\n",
     "import torchnet.meter as tnt\n",
     "t0 = time.time()\n",
-    "test_loader = imagenet_load_data(\"../../data.imagenet\", \n",
-    "                                 batch_size=512, \n",
+    "test_loader = imagenet_load_data(\"/home/cvds_lab/datasets/imagenet\", \n",
+    "                                 batch_size=64, \n",
```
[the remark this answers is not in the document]
guyjacob: Smaller batch size for machines with lesser resources

PR #149, opened 2019-02-07
Load different random subset of dataset on each epoch (CIFAR-10 / ImageNet only)

guyjacob:


examples/classifier_compression/parser.py
```diff
@@ -134,8 +145,15 @@ def linear_quant_mode_str(val_str):
     return parser
 
 
-def float_range(val_str):
+def float_range_exc_1(val_str):
```
nzmora: See if you like this code - I think it is more generic:
```
def float_range(minval=0, maxval=1):
    """Allows floats in [minval..maxval]"""
    def checker(val_str):
        val = float(val_str)
        if val < minval or val > maxval:
            raise argparse.ArgumentTypeError('Must be in range [{}..{}] (received {})'.format(minval, maxval, val_str))
        return val
    return checker
```

guyjacob: I saw your generic, and raised it some more generic :)

nzmora: I fold, then.

PR #150, opened 2019-02-09
Multiple breaking changes to distiller parser

barrh:
This branch contains several dependent proposals for small breaking updates, mainly to the parser.
I divided them for different sequential commits so we can still drop portion of them.

Although, it should work with torch=0.4.0, I would recommend this be merged towards distiller 0.3, just because it breaks current API.

### DIFF
--- a/distiller/apputils/checkpoint.py
+++ b/distiller/apputils/checkpoint.py
@@ -74,12 +74,17 @@ def save_checkpoint(epoch, arch, model, optimizer=None, scheduler=None,
 
 
 def load_checkpoint(model, chkpt_file, optimizer=None):
-    """Load a pytorch training checkpoint
+    """Load a pytorch training checkpoint.
+
+    Failure to retrieve optimizer from checkpoint will not
+    fail to call, it will return optimizer=None instead.
 
     Args:
         model: the pytorch model to which we will load the parameters
         chkpt_file: the checkpoint file
-        optimizer: the optimizer to which we will load the serialized state
+        optimizer: the optimizer to which we will load the serialized state.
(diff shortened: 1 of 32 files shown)


barrh:
I've noticed github doesn't guarantee the order of the commits in this view.
The correct order is: e066c6d, 1a42707, 0ab7a33, a21de3b, c1833d9

barrh:
Please **do not merge** it yet. I'll have it rebase ontop of my other standing PRs.

barrh:
I fpushed rebased version. The most notable change is deprecation warnings, so calling classifier.py with the old arguments will print a helpful warning and ignore it (excluding --cpu).
Please note that only the last 6 commits are proposed here, all the rest of the commits are from other standing PRs.

This patch fixes are:
* --print-freq(ency) will now set the _frequency_ of the prints to the logger. So for either cifar10, imagenet, or custom validation set, it will print exactly print_frequency times to pylogger.
* validation-size defaults to zero, so validation set == test set, unless requested otherwise.
Also add some deprecation warning for the old argument names.
* Rename workers/j to loaders, and set it to 5*gpus. Frankly, I can't think of a reason to specify this argument anymore. Calling with -j=1 will not fail the script, but it will be ignored with a deprecation warning. I set the #loaders=1 when deterministic is set, not sure if that's required.
* Rename --cpu to --use-cpu, and allow --gpus="" for the same purpose. Calling with --cpu will fail the run.

This patch does not fix any critical bug. Please close it if you do not wish to merge it.

barrh:
closed per @guyjacob request.

Issue #148, opened 2019-02-07
Run distiller on CPU


marcusvlc:
I'm trying to run the distiller on the CPU, but I'm getting a "keyerror" in the load_state_dict method. After debugging the code, I noticed that the error is in the state parameter of the method, which receives a dictionary with different name keys with each call, and when searching in that dictionary, the error happens. I already tried to execute the commands available in the README with the --cpu argument but did not get any positive results. What are the changes needed to run the master with cpu?

nzmora:
Hi @marcusvlc ,

Please send your exact command-line so we'll be able to recreate.
Thanks

marcusvlc:
The error is:

Log file for this run: /home/vinicius/testes/distiller/examples/classifier_compression/logs/2019.02.07-084425/2019.02.07-084425.log
==> using cifar10 dataset
=> creating resnet20_cifar model for CIFAR10

--------------------------------------------------------
Logging to TensorBoard - remember to execute the server:
> tensorboard --logdir='./logs'

=> loading checkpoint ../ssl/checkpoints/checkpoint_trained_dense.pth.tar
Checkpoint keys:
state_dict
	epoch
	arch
	best_top1
	compression_sched
	optimizer
   best top@1: 92.540

Log file for this run: /home/vinicius/testes/distiller/examples/classifier_compression/logs/2019.02.07-084425/2019.02.07-084425.log
Traceback (most recent call last):
  File "compress_classifier.py", line 723, in <module>
    main()
  File "compress_classifier.py", line 173, in main
    model, compression_scheduler, start_epoch = apputils.load_checkpoint(model, chkpt_file=args.resume)
  File "/home/vinicius/testes/distiller/apputils/checkpoint.py", line 97, in load_checkpoint
    compression_scheduler.load_state_dict(checkpoint['compression_sched'])
  File "/home/vinicius/testes/distiller/distiller/scheduler.py", line 213, in load_state_dict
    masker.mask = loaded_masks[name]
KeyError: 'layer1.0.bn2.bias'

nzmora:
Thanks - I've recreated.  
Sorry if this is delaying you - I'll try fixing in the next couple of days.
Neta

nzmora:
Hi @marcusvlc,

Please try this branch: https://github.com/NervanaSystems/distiller/tree/issue_148

Thanks!
Neta

PR #152, opened 2019-02-11
Fix Issue #148 - load GPU-optimized models on the CPU

nzmora:
The root-cause of this issue is that DataParallel modules cannot execute on the CPU - 
on machines that have both CPUs and GPUs.
Therefore, we don’t use DataParallel for models loaded for the CPUs, but we do wrap
the models with DataParallel when loaded on the GPUs (to make them run faster).
In PyTorch, the names of module keys saved in a checkpoint file depend if the modules are wrapped
by a DataParallel module or not (e.g. "module.conv1" vs. "conv1"). Therefore, loading a checkpoint that was optimized on the GPU, onto a CPU (and vice-versa) will fail on the keys (KeyError).
This is all PyTorch and despite the community asking for a fix -
e.g. pytorch/pytorch#7457 - it is still pending.

This commit contains code to catch key errors when loading a GPU-generated model
(i.e. with DataParallel) onto a CPU, and seamlessly convert the names of the keys.
We currently do not support the reverse direction (i.e. CPU-trained model loaded to the GPU).


nzmora:
@barrh 
I merged branch 'issue_148' into 'load_cp'.
It passes the tests - so I'll add you as reviewer and then we can merge this combined PR into 'master'.

PR #156, opened 2019-02-14
Store config files in logdir

barrh:
Store configuration files (args.compress) in the output directory, under 'configs' sub-directory.

### DIFF
--- a/apputils/execution_env.py
+++ b/apputils/execution_env.py
@@ -20,32 +20,36 @@ This is helpful if you want to recreate an experiment at a later time, or if
 you want to understand the environment in which you execute the training.
 """
 
-import sys
-import os
-import time
-import platform
+import contextlib
 import logging
 import logging.config
+import os
+import platform
+import shutil
+import sys
+import time
+
+from git import Repo, InvalidGitRepositoryError
 import numpy as np
 import torch
-from git import Repo, InvalidGitRepositoryError
-HAVE_LSB = True
 try:
     import lsb_release
+    HAVE_LSB = True
 except ImportError:
     HAVE_LSB = False
 
 logger = logging.getLogger("app_cfg")
 
 
-def log_execution_env_state(app_args, gitroot='.'):
+def log_execution_env_state(config_filename=None, logdir=None, gitroot='.'):
     """Log information about the execution environment.
 
     It is recommeneded to log this information so it can be used for referencing
     at a later time.
 
     Args:
-        app_args (dict): the command line arguments passed to the application
+        config_filename: filename to store in logdir, if logdir is set
+        logdir: log directory
         git_root: the path to the .git root directory
     """
 
@@ -81,7 +85,19 @@ def log_execution_env_state(app_args, gitroot='.'):
     logger.debug("PyTorch: %s", torch.__version__)
     logger.debug("Numpy: %s", np.__version__)
     log_git_state()
-    logger.debug("Command line: %s", " ".join(app_args))
+    logger.debug("Command line: %s", " ".join(sys.argv))
+
+    if (logdir is None) or (config_filename is None):
+        return
+    # clone configuration files to output directory
+    configs_dest = os.path.join(logdir, 'configs')
+    with contextlib.suppress(FileExistsError):
+        os.makedirs(configs_dest)
+    if os.path.exists(os.path.join(configs_dest, config_filename)):
+        logger.debug('{} already exists in logdir'.format(
+            os.path.basename(config_filename) or config_filename))
+    else:
+        shutil.copy(config_filename, configs_dest)
 
 
 def config_pylogger(log_cfg_file, experiment_name, output_dir='logs'):
--- a/examples/classifier_compression/compress_classifier.py
+++ b/examples/classifier_compression/compress_classifier.py
@@ -89,8 +89,7 @@ def main():
     global msglogger
 
     # Parse arguments
-    prsr = parser.get_parser()
-    args = prsr.parse_args()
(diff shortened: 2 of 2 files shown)


nzmora:
It's always advisable to write some explanation about PRs: what/why/how.
It makes it easier (and faster) for us to approach the code.
Thanks
Neta

barrh:
The ideal is: results will be easy to reproduce.
Storing results along with input files is a step in the right direction.
Another advantage is that it makes one's experiments much easier to track. (when conducting multiple experiments in parallel)

Therefore, I had copied the input file specified by args.compress to the logdir.
Ideally, all arguments that are required to reproduce an experiment would live in the same structure as the results.

apputils/execution_env.py
```diff
@@ -20,32 +20,37 @@
...
+def log_execution_env_state(app_argv, *config_fns, logdir=None, gitroot='.'):
     """Log information about the execution environment.
 
     It is recommeneded to log this information so it can be used for referencing
     at a later time.
 
     Args:
-        app_args (dict): the command line arguments passed to the application
+        app_argv (dict): the command line arguments passed to the application
+        config_fns: sequence of filenames to store in logdir, if set
```
nzmora: We use ```_fn``` to function variables, so this is confusing to me and I prefer you change it.  This is also true in the code below.

apputils/execution_env.py
```diff
@@ -81,7 +86,20 @@ def log_git_state():
...
-    logger.debug("Command line: %s", " ".join(app_args))
+    logger.debug("Command line: %s", " ".join(app_argv))
+
+    if logdir is None:
+        return
+    # clone configuration files to output directory
+    configs_dest = os.path.join(logdir, 'configs')
+    with contextlib.suppress(FileExistsError):
+        os.makedirs(configs_dest)
+    for fn in config_fns:
```
nzmora: ```for file_name in config_files:```

apputils/execution_env.py
```diff
@@ -81,7 +86,20 @@ def log_git_state():
...
+        if os.path.exists(os.path.join(configs_dest, fn)):
+            logger.debug('{} already exists in logdir'.format(
+                os.path.basename(fn) or fn))
+            continue
```
nzmora: please replace ```continue``` with ```else:```

apputils/execution_env.py
```diff
@@ -20,32 +20,37 @@
...
     import lsb_release
+    HAVE_LSB = True
 except ImportError:
     HAVE_LSB = False
 
 logger = logging.getLogger("app_cfg")
 
 
-def log_execution_env_state(app_args, gitroot='.'):
+def log_execution_env_state(app_argv, *config_fns, logdir=None, gitroot='.'):
```
nzmora: Why do we need *config_fns to be a list?  Is this future-proofing or do you use it?
More often than not, future-proofing leads to extra code that rots over time.  If you don't really pass a list of file names, then please simplify.

examples/classifier_compression/compress_classifier.py
```diff
@@ -89,16 +89,20 @@ def main():
...
 
     # Log various details about the execution environment.  It is sometimes useful
     # to refer to past experiment executions and this information may be useful.
-    apputils.log_execution_env_state(sys.argv, gitroot=module_path)
+    if args.compress is None:
+        apputils.log_execution_env_state(sys.argv, gitroot=module_path)
+    else:
+        apputils.log_execution_env_state(sys.argv, args.compress,
+            logdir=msglogger.logdir, gitroot=module_path,
+        )
```
nzmora: * please remove the new-line here (i.e. place the closing parenthesis on the previous line).
* in fact, we don't need the ```else``` clause, do we?  You should handle the case of ```args.compress== None``` in ```apputils.log_execution_env_state```, not here.

apputils/execution_env.py
barrh: config_filename (singular) it is.

apputils/execution_env.py
barrh: I see nothing wrong with future proofing.
going over a loop of no elements in python is usually cleaner than checking if argument is none or not.

examples/classifier_compression/compress_classifier.py
barrh: It's correct either way, according to Python's styling guidelines.

nzmora:
Are you rebasing the code?  Please don't do that next time because it turns all comments to "Outdated" and I can't see what you changed.

barrh:
@nzmora I f-pushed with the requested changes.
Also edited the description, and retested.

apputils/execution_env.py
```diff
@@ -20,32 +20,37 @@
...
     import lsb_release
+    HAVE_LSB = True
 except ImportError:
     HAVE_LSB = False
 
 logger = logging.getLogger("app_cfg")
 
 
-def log_execution_env_state(app_args, gitroot='.'):
+def log_execution_env_state(app_argv=None, config_filename=None, logdir=None, gitroot='.'):
```
nzmora: Why assign a default value to ```app_argv```?  It should always be available in my opinion.

barrh: In the current context, you are correct. Change it if it bothers you.
I figured it's perfectly useful to call "log_execution_env_state()" with no arguments at all, so why not allow it.

barrh:
there's however, a small issue with the description of the function... lemme f-push correction.

barrh:
@nzmora I moved the call to sys.argv into the function.

PR #160, opened 2019-02-18
Fix if_file_exists check on config file

barrh:
small fix. not very important.
The destination of the os.path.exists() call was incorrect, so the result was inconsistent.

### DIFF
--- a/distiller/apputils/execution_env.py
+++ b/distiller/apputils/execution_env.py
@@ -41,14 +41,16 @@ except ImportError:
 logger = logging.getLogger("app_cfg")
 
 
-def log_execution_env_state(config_filename=None, logdir=None, gitroot='.'):
+def log_execution_env_state(config_path=None, logdir=None, gitroot='.'):
     """Log information about the execution environment.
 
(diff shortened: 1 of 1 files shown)


apputils/execution_env.py
```diff
@@ -41,14 +41,14 @@
 logger = logging.getLogger("app_cfg")
 
 
-def log_execution_env_state(config_filename=None, logdir=None, gitroot='.'):
+def log_execution_env_state(config_path=None, logdir=None, gitroot='.'):
     """Log information about the execution environment.
 
     It is recommeneded to log this information so it can be used for referencing
```
nzmora: Please consider adding:
File 'config_path' will be copied to directory 'logdir'.  A common use-case is passing the path to a (compression) schedule YAML file.  Storing a copy of the schedule file with the experiment logs helps reproduce experiments because the schedule file contains important configuration information.

barrh: Edited the description.

barrh:
@nzmora I fpushed another commit so to wrap the copy() call with try-except.
Please merge this commit, unless there're new blockers.

PR #169, opened 2019-02-23
Resume optimizer

barrh:
This fixes long-standing bug in distiller with resuming from checkpoint.
Before this patch, the optimizer field in the checkpoint was ignored.

Although it corrects an erroneous behavior of the 'resume' flag, it _is_ a behavioral change of dominant functionality.

**A few thoughts before merging into master:**
* Documentation: This patch does not contain required changes to Distiller online documentation.
* Insufficient tests: None of current tests caught that bug, nor I address this matter in my patch.
I tested it briefly with the attached steplr lr_sched recipe, and enjoyed using it, but don't trust me on this - test for yourself. I have no idea how it plays with quantization, gpu<-->cpu transitions, or non-sgd optimizers.
* examples change: In the patch I added a flag that effectively reverts to the old incorrect behavior, but I didn't revise the examples in distiller.
1. I assume there's no example-file or notebook in Distiller that intentionally tries to take advantage of this bug.
2. with post-training pruning in mind - when one resumes local distiller checkpoint in order to prune it, it contains the old optimizer, and therefore what he probably expects is old behavior. This can be achieved with the 'reset-lr'/'reset-optimizer' flag. Using '--pretrained' without resume is not affected.

Please review, validate, test, and then merge with caution.

### DIFF
--- a/examples/classifier_compression/compress_classifier.py
+++ b/examples/classifier_compression/compress_classifier.py
@@ -164,8 +164,9 @@ def main():
     # We can optionally resume from a checkpoint
     optimizer = None
     if args.resume:
-        # initiate SGD with dummy lr
-        optimizer = torch.optim.SGD(model.parameters(), lr=0.36787944117)
+        if not args.reset_optimizer:
+            # initiate SGD with dummy lr
+            optimizer = torch.optim.SGD(model.parameters(), lr=0.36787944117)
         model, compression_scheduler, optimizer, start_epoch = apputils.load_checkpoint(
             model, args.resume, optimizer=optimizer)
         model.to(args.device)
@@ -236,7 +237,7 @@ def main():
         # The main use-case for this sample application is CNN compression. Compression
         # requires a compression schedule configuration file in YAML.
         compression_scheduler = distiller.file_config(model, optimizer, args.compress, compression_scheduler,
-            (start_epoch-1) if args.resume else None)
+            (start_epoch-1) if (args.resume and not args.reset_optimizer) else None)
         # Model is re-transferred to GPU in case parameters were added (e.g. PACTQuantizer)
         model.to(args.device)
     elif compression_scheduler is None:
--- a/examples/classifier_compression/compress_classifier.py
+++ b/examples/classifier_compression/compress_classifier.py
@@ -163,12 +163,12 @@ def main():
 
     # We can optionally resume from a checkpoint
     optimizer = None
-    if args.resume:
-        if not args.reset_optimizer:
+    if args.resume or args.load_state_dict:
+        if args.resume and not args.reset_optimizer:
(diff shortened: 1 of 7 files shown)


examples/classifier_compression/compress_classifier.py
```diff
@@ -162,18 +162,37 @@ def main():
...
+        # check if user has tried to set optimizer arguments
+        # if so, ignore arguments with a warning.
+        optimizer_group_args = ['lr', 'learning-rate', 'momentum', 'weight-decay', 'wd']
+        user_optim_args = [
+                x for x in optimizer_group_args for arg in sys.argv
+                if arg.startswith('--' + x)
+            ]
+        if user_optim_args:
+            msglogger.warning('{} optimizer arguments are ignored.'.format(user_optim_args))
+            msglogger.info('setting optimizer arguments when optimizer is resumed '
```
guyjacob: I appreciate the diligence of performing this check :) But, it's going to be a pain to maintain if we decide to try and support other optimizers with other arguments, no? So I don't mind this as it is for now, but I also think just issuing a "blanket" warning whenever an optimizer was loaded from a checkpoint would be good enough here (and would reduce a little bit of bloat from the code).

examples/classifier_compression/parser.py
```diff
@@ -40,16 +40,26 @@ def get_parser():
...
     parser.add_argument('-b', '--batch-size', default=256, type=int,
                         metavar='N', help='mini-batch size (default: 256)')
-    parser.add_argument('--lr', '--learning-rate', default=0.1, type=float,
-                        metavar='LR', help='initial learning rate')
-    parser.add_argument('--momentum', default=0.9, type=float, metavar='M',
-                        help='momentum')
-    parser.add_argument('--weight-decay', '--wd', default=1e-4, type=float,
-                        metavar='W', help='weight decay (default: 1e-4)')
+
+    optimizer_args = parser.add_argument_group('optimizer_arguments')
```
guyjacob: The string you pass here is printed when you run with `--help`, so it should be more human-readable:
```suggestion
    optimizer_args = parser.add_argument_group('Optimizer Arguments')
```

tests/full_flow_tests/steplr_training_resnet20_cifar.yaml
```diff
@@ -0,0 +1,14 @@
+# python3 compress_classifier.py  --lr=0.1 --arch=resnet20_cifar ../../../data.cifar10 --compress=../../tests/full_flow_tests/steplr_training_resnet20_cifar.yaml
```
guyjacob: You wrote that you used this to test, but I don't see you're pushing any actual tests with it. So either add tests, or remove the file.

barrh:
I won't pursue this. abandoned.

PR #173, opened 2019-02-25
ACIQ clipping in post-training quantization

ynahshan:
PostTraining:
-- Add aciq implementation and use in dynamic mode
-- Fix scheduler bug (code from Guy) can be omitted if already in master


nzmora:
Hi @levzlotnik 
Is it ready for review?

distiller/quantization/q_utils.py
```diff
@@ -175,6 +176,79 @@ def approx_scale_as_mult_and_shift(fp32_scale, mult_bits, limit=False):
...
+    def __call__(self, t, across_dim=None):
+        if self.clip_type == AciqClipper.AciqClippingType.Laplace:
+            alpha = super(AciqAsymmetricClipper, self).get_alpha_laplace(t, across_dim, self.num_bits, half_range=self.half_range)
+        else:
+            alpha = super(AciqAsymmetricClipper, self).get_alpha_gauss(t, across_dim, self.num_bits, half_range=self.half_range)
+        if isinstance(t, dict):
+            mean, min_val = torch.tensor(t['mean']), torch.tensor(t['avg_min'])
+        else:
+            mean = t.mean()
+            min_val = get_tensor_min_max(t, across_dim)[0].mean()
```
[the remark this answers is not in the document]
levzlotnik: `get_tensor_avg_min_max` calculates means of both `min` and `max`, while we only need `min`. 
no need for unnecessary calculations.

distiller/quantization/range_linear.py
```diff
@@ -953,7 +967,7 @@ def __init__(self, model, bits_activations=8, bits_parameters=8, bits_accum=32,
 
         def replace_param_layer(module, name, qbits_map, per_channel_wts=per_channel_wts,
                                 mode=mode, fp16=fp16, scale_approx_mult_bits=scale_approx_mult_bits,
-                                clip_acts=clip_acts, clip_n_stds=clip_n_stds):
+                                clip_acts=clip_acts, clip_half_range=False, clip_n_stds=clip_n_stds):
```
guyjacob: Passing `clip_half_range` as a parameter to this function - I assume you intend for this to be set via overrides? Why not just check for this in the function, instead of putting the burden on the user? Something like:
```python
succ = self.adjacency_map[name].successors
clip_half_range = len(succs == 1) and succs[0].type == 'Relu'
```
(Wrote this of the top off my head, so might have gotten the attribute name wrong...)
Isn't a check like this enough? Unless you envision a usage for half-range other than followed-by-relu.

guyjacob: This applies to `replace_non_param_layer` as well, of course.

guyjacob: We'll merge it as is for the moment, and try the "auto-detection" separately.

guyjacob:
I'm merging this. Still TODO:
* Update documentation and examples
* Try "auto-detection" of half-range within `PostTrainLinearQuantizer`

PR #178, opened 2019-02-27
Extend Quantizer Functionality

levzlotnik:
NOTE - this is a re-opened pull request.  
Added `overrides` argument to Quantizer ctor, to allow overriding parameters other than `bits_override`.  
Added `QBit.bias` field to allow quantizing bias.


distiller/quantization/quantizer.py
```diff
@@ -91,18 +91,20 @@ class Quantizer(object):
...
     """
-    def __init__(self, model, optimizer=None, bits_activations=None, bits_weights=None, bits_overrides=None,
-                 quantize_bias=False, train_with_fp_copy=False):
-        if bits_overrides is None:
-            bits_overrides = OrderedDict()
-        if not isinstance(bits_overrides, OrderedDict):
-            raise TypeError('bits_overrides must be an instance of collections.OrderedDict or None')
+    def __init__(self, model, optimizer=None,
+                 bits_activations=None, bits_weights=None, bits_bias=None,
+                 overrides=None,
```
guyjacob: It's fine that you defined a new parameter named just `overrides`, indeed that conveys better what that parameter is supposed to be doing now.
But, there's a lot of existing code and, more importantly, existing YAML files that have `bits_overrides` defined in them. In order not make the transition smoother and not break everything, let's also keep the older `bits_overrides` parameter as well, with a comment that it's deprecated but kept for backwards compatibility. See [this](https://docs.scipy.org/doc/numpy-1.15.0/reference/generated/numpy.histogram.html) for example (the normed and density parameters).

distiller/quantization/quantizer.py
```diff
@@ -91,18 +91,20 @@ class Quantizer(object):
...
+        if not isinstance(overrides, OrderedDict):
+            raise TypeError('overrides must be an instance of collections.OrderedDict or None')
 
         if train_with_fp_copy and optimizer is None:
             raise ValueError('optimizer cannot be None when train_with_fp_copy is True')
 
-        self.default_qbits = QBits(acts=bits_activations, wts=bits_weights)
-        self.quantize_bias = quantize_bias
+        self.default_qbits = QBits(acts=bits_activations, wts=bits_weights, bias=bits_bias)
+        bits_overrides = overrides.get('bits', OrderedDict())
```
guyjacob: Again, I'm worried about backwards compatibility. All existing YAML files (the ones in our examples AND whatever users defined on their own) will have the overrides defined similar to:
```yaml
bits_overrides:
  conv1:
    wts: 4
    acts: 4
...
```
But here you added a "sub-dict" named `bits` for the bit-widths. Which makes some sense if we were doing this from scratch, but now will break all existing inputs.
So at least for now I prefer to keep, or at least still support, the existing structure of the overrides dict. So the bit-widths and the other parameters will be at the same level. That means we need some extra logic to extract the bit-widths. But, again, I think it's worth it.

levzlotnik:
@guyjacob please check out the fixes

levzlotnik:
@guyjacob Please review the new changes

levzlotnik:
@guyjacob @nzmora Can we merge into master?

PR #179, opened 2019-02-27
Resume training steps

barrh:
I address in this patch several issues/bugs:
 * Fix weights distribution histogram. (It's broken.)
 * Fix training steps: the former behavior assumed batch size is a constant
    Before the patch, when resumed training with different batch size, the training step was skewed
 * Fix typo in tflogger path (this will have visual effect on comparing new log files with the old!)
 * Fix utils.log_training_progress. It was misused, and therefore, bug-prone.
    I renamed the function, so it makes the refactoring safer to commit.
    I limited the use of this function to where I saw fit. Although both serve similar purpose, Pylogger and TFlogger are very different beasts, and require different parameters more often than not.

I tested it, and enjoyed using it, but I might have overlooked many scenarios, especially in ADC.py and main.py areas which I'm hardly familiar with.

This branch is dependent on 2 unmerged PRs: #169  and #132

### DIFF
--- a/distiller/apputils/checkpoint.py
+++ b/distiller/apputils/checkpoint.py
@@ -74,12 +74,17 @@ def save_checkpoint(epoch, arch, model, optimizer=None, scheduler=None,
 
 
 def load_checkpoint(model, chkpt_file, optimizer=None):
-    """Load a pytorch training checkpoint
+    """Load a pytorch training checkpoint.
+
+    Failure to retrieve optimizer from checkpoint will not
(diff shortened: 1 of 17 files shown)


barrh:
abandoned

nzmora:
Commits a7cb1561622bb43856ab2b6633db0e2a18fc05c3, 5533aa1128bf9cdbd1ef8ee4870d0d02db6fff62, and 4da83eb3f10c9604c72f0a00913266561e2d16ca should be merged to 'master'.

PR #182, opened 2019-03-06
Load optimizer

barrh:
[ resubmitting #169 ]

This fixes long-standing bug in distiller with resuming from checkpoint.
Before this patch, the optimizer field in the checkpoint was ignored.

Although it corrects an erroneous behavior of the 'resume' flag, it is a behavioral change of dominant functionality.

A few thoughts before merging into master:

Documentation: This patch does not contain required changes to Distiller online documentation.
Insufficient tests: None of current tests caught that bug, nor I address this matter in my patch.
I tested it briefly with the attached steplr lr_sched recipe, and enjoyed using it, but don't trust me on this - test for yourself. I have no idea how it plays with quantization, gpu<-->cpu transitions, or non-sgd optimizers.
examples change: In the patch I added a flag that effectively reverts to the old incorrect behavior, but I didn't revise the examples in distiller.
I assume there's no example-file or notebook in Distiller that intentionally tries to take advantage of this bug.
with post-training pruning in mind - when one resumes local distiller checkpoint in order to prune it, it contains the old optimizer, and therefore what he probably expects is old behavior. This can be achieved with the 'reset-lr'/'reset-optimizer' flag. Using '--pretrained' without resume is not affected.

Please review, validate, test, and then merge with caution.

### DIFF
--- a/distiller/policy.py
+++ b/distiller/policy.py
@@ -21,6 +21,7 @@
 - LRPolicy: learning-rate decay scheduling
 """
 import torch
+import torch.optim.lr_scheduler
 from collections import namedtuple
 
 import logging
@@ -210,7 +211,11 @@ class LRPolicy(ScheduledTrainingPolicy):
         self.lr_scheduler = lr_scheduler
 
     def on_epoch_begin(self, model, zeros_mask_dict, meta, **kwargs):
-        self.lr_scheduler.step(**kwargs)
+        if isinstance(self.lr_scheduler, torch.optim.lr_scheduler.ReduceLROnPlateau):
+            # Note: ReduceLROnPlateau doesn't inherit from _LRScheduler
+            self.lr_scheduler.step(**kwargs)
+        else:
+            self.lr_scheduler.step(epoch=kwargs.get('epoch', None))
 
 
 class QuantizationPolicy(ScheduledTrainingPolicy):
--- a/distiller/policy.py
+++ b/distiller/policy.py
@@ -42,7 +42,7 @@ class ScheduledTrainingPolicy(object):
         self.classes = classes
         self.layers = layers
 
-    def on_epoch_begin(self, model, zeros_mask_dict, meta):
+    def on_epoch_begin(self, model, zeros_mask_dict, meta, **kwargs):
         """A new epcoh is about to begin"""
         pass
 
@@ -115,7 +115,7 @@ class PruningPolicy(ScheduledTrainingPolicy):
         self.mini_batch_id = 0          # The ID of the mini_batch within the present epoch
         self.global_mini_batch_id = 0   # The ID of the mini_batch within the present training session
 
-    def on_epoch_begin(self, model, zeros_mask_dict, meta):
+    def on_epoch_begin(self, model, zeros_mask_dict, meta, **kwargs):
         msglogger.debug("Pruner {} is about to prune".format(self.pruner.name))
         self.mini_batch_id = 0
         self.is_last_epoch = meta['current_epoch'] == (meta['ending_epoch'] - 1)
@@ -169,7 +169,7 @@ class RegularizationPolicy(ScheduledTrainingPolicy):
         self.keep_mask = keep_mask
         self.is_last_epoch = False
 
-    def on_epoch_begin(self, model, zeros_mask_dict, meta):
+    def on_epoch_begin(self, model, zeros_mask_dict, meta, **kwargs):
         self.is_last_epoch = meta['current_epoch'] == (meta['ending_epoch'] - 1)
 
     def before_backward_pass(self, model, epoch, minibatch_id, minibatches_per_epoch, loss,
@@ -202,15 +202,15 @@ class RegularizationPolicy(ScheduledTrainingPolicy):
 
 
 class LRPolicy(ScheduledTrainingPolicy):
-    """ Learning-rate decay scheduling policy.
+    """Learning-rate decay scheduling policy.
 
     """
     def __init__(self, lr_scheduler):
         super(LRPolicy, self).__init__()
         self.lr_scheduler = lr_scheduler
 
-    def on_epoch_begin(self, model, zeros_mask_dict, meta):
-        self.lr_scheduler.step()
+    def on_epoch_begin(self, model, zeros_mask_dict, meta, **kwargs):
+        self.lr_scheduler.step(**kwargs)
 
 
 class QuantizationPolicy(ScheduledTrainingPolicy):
--- a/distiller/policy.py
+++ b/distiller/policy.py
@@ -213,9 +213,9 @@ class LRPolicy(ScheduledTrainingPolicy):
     def on_epoch_begin(self, model, zeros_mask_dict, meta, **kwargs):
         if isinstance(self.lr_scheduler, torch.optim.lr_scheduler.ReduceLROnPlateau):
             # Note: ReduceLROnPlateau doesn't inherit from _LRScheduler
-            self.lr_scheduler.step(**kwargs)
+            self.lr_scheduler.step(kwargs['metrics'], epoch=meta['current_epoch'])
         else:
-            self.lr_scheduler.step(epoch=kwargs.get('epoch', None))
+            self.lr_scheduler.step(epoch=meta['current_epoch'])
 
 
 class QuantizationPolicy(ScheduledTrainingPolicy):
--- a/distiller/scheduler.py
+++ b/distiller/scheduler.py
@@ -111,7 +111,7 @@ class CompressionScheduler(object):
             meta = self.sched_metadata[policy]
             meta['current_epoch'] = epoch
             policy.on_epoch_begin(self.model, self.zeros_mask_dict, meta,
-                                  epoch=epoch, **kwargs)
+                                  **kwargs)
 
     def on_minibatch_begin(self, epoch, minibatch_id, minibatches_per_epoch, optimizer=None):
         if epoch in self.policies:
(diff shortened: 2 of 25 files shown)


distiller/policy.py
```diff
@@ -202,15 +203,19 @@ def on_minibatch_end(self, model, epoch, minibatch_id, minibatches_per_epoch, ze
...
     def __init__(self, lr_scheduler):
         super(LRPolicy, self).__init__()
         self.lr_scheduler = lr_scheduler
 
-    def on_epoch_begin(self, model, zeros_mask_dict, meta):
-        self.lr_scheduler.step()
+    def on_epoch_begin(self, model, zeros_mask_dict, meta, **kwargs):
+        if isinstance(self.lr_scheduler, torch.optim.lr_scheduler.ReduceLROnPlateau):
+            # Note: ReduceLROnPlateau doesn't inherit from _LRScheduler
+            self.lr_scheduler.step(**kwargs)
```
guyjacob: Elsewhere we use `meta['current_epoch]` to get the epoch number, let's be consistent with that. The other 2 comments are related to this as well.
```suggestion
            self.lr_scheduler.step(kwargs['metrics'], meta['current_epoch'])
```

distiller/policy.py
```diff
@@ -202,15 +203,19 @@ def on_minibatch_end(self, model, epoch, minibatch_id, minibatches_per_epoch, ze
...
+        else:
+            self.lr_scheduler.step(epoch=kwargs.get('epoch', None))
```
guyjacob: Get epoch from `meta` dict.
```suggestion
            self.lr_scheduler.step(meta['current_epoch'])
```

distiller/scheduler.py
```diff
@@ -106,12 +106,12 @@ def add_policy(self, policy, epochs=None, starting_epoch=0, ending_epoch=1, freq
...
-            for policy in self.policies[epoch]:
-                meta = self.sched_metadata[policy]
-                meta['current_epoch'] = epoch
-                policy.on_epoch_begin(self.model, self.zeros_mask_dict, meta)
+    def on_epoch_begin(self, epoch, optimizer=None, **kwargs):
+        for policy in self.policies.get(epoch, list()):
+            meta = self.sched_metadata[policy]
+            meta['current_epoch'] = epoch
+            policy.on_epoch_begin(self.model, self.zeros_mask_dict, meta,
+                                  epoch=epoch, **kwargs)
```
guyjacob: Epoch number is in `meta`, no need to pass it separately.
```suggestion
                                  **kwargs)
```

examples/classifier_compression/compress_classifier.py
```diff
@@ -233,7 +243,7 @@ def main():
     if args.kd_teacher:
         teacher = create_model(args.kd_pretrained, args.dataset, args.kd_teacher, device_ids=args.gpus)
         if args.kd_resume:
-            teacher, _, _ = apputils.load_checkpoint(teacher, chkpt_file=args.kd_resume)
+            teacher = apputils.load_checkpoint(teacher, chkpt_file=args.kd_resume)[0]
```
nzmora: Personally, I think the code as originally written is clearer (and I suspect, more "pythonic")

barrh: You were commenting an outdated commit.

examples/classifier_compression/compress_classifier.py
```diff
@@ -159,18 +159,27 @@ def main():
         msglogger.info('=> using early-exit threshold values of %s', args.earlyexit_thresholds)
 
     # We can optionally resume from a checkpoint
+    optimizer = None
     if args.resume:
-        model, compression_scheduler, start_epoch = apputils.load_checkpoint(model, chkpt_file=args.resume)
+        # initiate SGD with dummy lr
+        optimizer = torch.optim.SGD(model.parameters(), lr=0.36787944117)
```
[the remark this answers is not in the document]
barrh: This is an outdated snippet.

barrh:
@nzmora Some of your comments were addressing outdated code (with classifier.py specifically). I marked those comments resolved for now.
You are welcome to send another review.

examples/classifier_compression/compress_classifier.py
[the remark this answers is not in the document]
guyjacob: The code indeed changed since the snippet Neta is commenting on, but the idea he is proposing is still valid.
It does mean that we need to add another parameter to `load_checkpoint` - the optimizer type. It can have a default of None, in which case the optimizer isn't loaded from the checkpoint, maintaining the old behavior.
I'm in favor of this.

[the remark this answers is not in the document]
nzmora: We can't change the behavior of code that has been out there for almost a year.  And certainly when the reasons are not compelling enough (we can support old and new behavior w/o much effort).

barrh:
Opened related Issue #187 to this PR. We might need to move the to() call into load_checkpoint, in order for the returned optimizer to be correct. What do you think?

nzmora:
@barrh our current flow is contrary to PyTorch's recommendation 
(https://pytorch.org/docs/stable/optim.html):
> If you need to move a model to GPU via .cuda(), please do so before constructing optimizers for it. Parameters of a model after .cuda() will be different objects with those before the call.
In general, you should make sure that optimized parameters live in consistent locations when optimizers are constructed and used.

So if we first move the model (```model.to(args.device)```), then load_checkpoint can use ```distiller.model_device(model)``` to infer the device placement for the optimizer.

barrh:
Similarly to optimizer, is there any reason not to store the model parameters in the checkpoint as well (and generate model from checkpoint)?

nzmora:
What do you refer to by "model parameters"?

barrh:
Model parameters is everything required to regenerate the model: state_dict, type, layer naming format?
If load_checkpoint recreates the optimizer, it would make sense to recreate the model as well. Plus, it's useful, because no longer one has to have prior knowledge of the content of a checkpoint in order to reload it. Path to a valid checkpoint is all you will ever going to need.

nzmora:
We save the state_dict.
Layer naming depends on the device.
Format? I assume you mean the arch/architecture of the network.  That is something that would be beneficial (to be able to use ```inspect_ckpt.py``` to describe the contents).
But I don't know of other parameters that would be helpful.
To instantiate the model you still need the model code, and this can change (especially for non-TorchVision models).  
Are there specific parameters you know of that would be helpful to store?

examples/classifier_compression/compress_classifier.py
```diff
@@ -158,19 +161,36 @@ def main():
...
+        model, compression_scheduler, optimizer, start_epoch = apputils.load_checkpoint(
+            model, args.resumed_checkpoint_path, model_device=args.device)
+    elif args.load_model_path:
+        model = apputils.load_lean_checkpoint(model, args.load_model_path,
+                                              model_device=args.device)
+    if args.reset_optimizer:
+        start_epoch = 0
+        if optimizer is not None:
+            optimizer = None
+            msglogger.info('Overriding resumed optimizer')
```
guyjacob: Just to make it clearer and stand out a bit more in the log:
```suggestion
            msglogger.info()
            msglogger.info('reset_optimizer flag set: Overriding resumed optimizer and resetting epoch count to 0')
```

examples/classifier_compression/parser.py
```diff
@@ -36,24 +36,40 @@ def get_parser():
...
+    optimizer_args.add_argument('--weight-decay', '--wd', default=1e-4, type=float,
+                    metavar='W', help='weight decay (default: 1e-4)')
+
     parser.add_argument('--print-freq', '-p', default=10, type=int,
                         metavar='N', help='print frequency (default: 10)')
-    parser.add_argument('--resume', default='', type=str, metavar='PATH',
-                        help='path to latest checkpoint (default: none)')
+
+    load_checkpoint_group = parser.add_argument_group('Resuming arguments')
+    load_checkpoint_group_exc = parser.add_mutually_exclusive_group()
```
guyjacob:
```suggestion
    load_checkpoint_group_exc = load_checkpoint_group.add_mutually_exclusive_group()
```
This way the mutually exclusive group also appears under the "Resuming arguments" header when help is printed.

examples/classifier_compression/compress_classifier.py
```diff
@@ -158,19 +161,36 @@ def main():
     if args.earlyexit_thresholds:
         msglogger.info('=> using early-exit threshold values of %s', args.earlyexit_thresholds)
 
-    # We can optionally resume from a checkpoint
-    if args.resume:
-        model, compression_scheduler, start_epoch = apputils.load_checkpoint(model, chkpt_file=args.resume)
-        model.to(args.device)
+    # TODO(barrh): args.deprecated_resume is deprecated since v0.3
```
guyjacob: Seeing as v0.3 is already tagged, this is a bit mis-leading. We'll likely bump the version number with this change and maybe a couple more. Not sure if we'll call it 0.3.1 or 0.4. So for now could you remove the "since v0.3"?

examples/classifier_compression/parser.py
```diff
@@ -36,24 +36,40 @@ def get_parser():
...
+    # TODO(barrh): args.deprecated_resume is deprecated since v0.3
```
guyjacob: Same comment as above re "since v0.3".

examples/classifier_compression/parser.py
barrh: @guyjacob  thanks! I missed that one.

PR #184, opened 2019-03-10
Integrate Cadene pretrainedmodels [fixes #142]

### LINKED ISSUE #142: Integrate Cadene pre-trained models
[Cadene](https://github.com/Cadene/pretrained-models.pytorch) is a great repository of pre-trained Convnets.  It's probably the largest single source of pre-trained models available for PyTorch. 

Many people are probably using Cadene and therefore it would be helpful to offer a simple bridge to Distiller, explaining how to use Cadene models from [compress_classifier.py](https://github.com/NervanaSystems/distiller/blob/master/examples/classifier_compression/compress_classifier.py).

Cheers
Neta

barrh:
Edited create_model() to import models from pretrainedmodels.
Added tests to the same function.

### DIFF
--- a/distiller/models/__init__.py
+++ b/distiller/models/__init__.py
@@ -49,18 +49,16 @@ def create_model(pretrained, dataset, arch, parallel=True, device_ids=None):
     """Create a pytorch model based on the model architecture and dataset
 
     Args:
-        pretrained: True is you wish to load a pretrained model.  Only torchvision models
-          have a pretrained model.
+        pretrained [boolean]: True is you wish to load a pretrained model.
+            Some models do not have a pretrained version.
         dataset: dataset name (only 'imagenet' and 'cifar10' are supported)
         arch: architecture name
-        parallel (boolean): if set, use torch.nn.DataParallel
+        parallel [boolean]: if set, use torch.nn.DataParallel
         device_ids: Devices on which model should be created -
             None - GPU if available, otherwise CPU
             -1 - CPU
             >=0 - GPU device IDs
     """
-    msglogger.info('==> using %s dataset' % dataset)
-
     model = None
     dataset = dataset.lower()
     if dataset == 'imagenet':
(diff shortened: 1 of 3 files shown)


distiller/models/__init__.py
```diff
@@ -20,6 +20,7 @@
 import torchvision.models as torch_models
 from . import cifar10 as cifar10_models
 from . import imagenet as imagenet_extra_models
+import pretrainedmodels
```
nzmora: I can't believe they named their package with such a generic name, and got away with it!

distiller/models/__init__.py
```diff
@@ -62,29 +62,37 @@ def create_model(pretrained, dataset, arch, parallel=True, device_ids=None):
...
+            model = imagenet_extra_models.__dict__[arch]()
         elif arch in pretrainedmodels.model_names:
             model = pretrainedmodels.__dict__[arch](
                         pretrained=(dataset if pretrained else None))
         else:
-            assert not pretrained, "Model %s (ImageNet) does not have a pretrained model" % arch
-            model = imagenet_extra_models.__dict__[arch]()
+            error_message = ''
+            if arch not in IMAGENET_MODEL_NAMES:
+                error_message = "Model %s is not supported for dataset %s" % (arch, 'ImageNet')
```
nzmora: It's not that important, but we know that the dataset is ImageNet, so why pass it as a parameter to the string constructor (as in line 81)?

barrh: I agree. I guess I didn't give it much thought. I also think the info message on line $62 is probably redundant.
Ping me tomorrow if you want me refine to this.

barrh:
Regarding your warning message when working when pretrainedmodels, please note I couldn't load 'nasnetalarge' with this patch. Moreover, the pretrained 'nasnetamoblie' (which does load) doesn't perform as expected out of the box.
I didn't have much time to look into that, but it looks like these models are WIP.

Issue #195, opened 2019-03-19
Accuracy mismatch using classifier.py - test() != validate()


barrh:
I executed resnet50 with pruning, validation set and test set are the same. But for some reason, for the same weights, validate() and test() returned (similar but) different statistics.

On the last epoch, these are the results I got:
--- validate (epoch=122)-----------
50000 samples (440 per mini-batch)
Epoch: [122][   57/  114]    Loss 0.966480    Top1 76.120415    Top5 92.838915
Epoch: [122][  114/  114]    Loss 0.960678    Top1 76.186000    Top5 92.882000
==> Top1: 76.186    Top5: 92.882    Loss: 0.961

--- test ---------------------
50000 samples (440 per mini-batch)
Test: [   57/  114]    Loss 0.963796    Top1 76.251994    Top5 92.799043
Test: [  114/  114]    Loss 0.962186    Top1 76.186000    Top5 92.882000
==> Top1: 76.186    Top5: 92.882    Loss: 0.962

[checkpoint at: 2019.03.17-121241]

nzmora:
Hi @barrh 

Because you expect that the results for test and validate to be equal, I assume that you ran ```compress_classifier.py``` with ```--vs=0```.

I'm pretty sure you see these insignificant differences due to the way we accumulate the statistics.  When you use multiple worker threads for the ```dataloader```, the inputs are processed in a non-deterministic order and that may slightly impact the intermediate  statistics.  
I.e. this:
```Epoch: [122][ 57/ 114] Loss 0.966480 Top1 76.120415 Top5 92.838915```
vs. this:
```Test: [ 57/ 114] Loss 0.963796 Top1 76.251994 Top5 92.799043```

The final statistics are the same except for the thousands digit in the loss (0.961 vs 0.962).  I don't have an immediate explanation, but I'm sure it's also due to the different order inputs are processed.

Cheers
Netsa

barrh:
Hey,
Thanks for your quick answer.

I close this issue because the difference is indeed insignificant, and evaluating the model with --deterministic provides a consistent result.

guyjacob:
Just to clarify, it's not related to multiple workers. The way the data loader is implemented now, it randomizes the order of images in the set each time it's iterated over, regardless of the number of workers. This was done to enable using only a subset of the test/validation set, where we want the subset to be representative AND we want a different subset each time (useful in auto-compression, for example).

[...]

PR #197, opened 2019-03-21
Deterministic test loader

barrh:
Pytorch current optimization for dataset loading does not guarantee order, this results in inconsistent accuracy results as I have reported in issue #195 .

Since the expected behavior for a given test-set and weights is to produce consistent results, this patch addresses that essentially by **(1)** passing over Pytorch's optimization for the test loader. (**THIS CHANGE HAS BEEN REVERTED** it will only behave this way during evaluation)
This patch also **(2)** passes the responsibility to set the environment parameters from the user to Distiller.
**(3)** It sets all the required cudnn parameters that Pytorch requires to guarantee determinism.

Please note that this patch will slow down the test loader.


barrh:
requires rebasing and clean-up.

Issue #198, opened 2019-03-25
Binarization with DoReFa-Net


tacker-oh:
This issue is about [DorefaParamsBinarizationSTE](https://github.com/NervanaSystems/distiller/blob/master/distiller/quantization/clipped_linear.py#L139).  

[`torch.sign()`](https://pytorch.org/docs/stable/torch.html#torch.sign) returns one of three possible values: {-1, 0, 1} .   I think this isn't binarization.   
https://github.com/NervanaSystems/distiller/blob/b8fc508528bd3f3d73da7bd62bafa4a5afc9001a/distiller/quantization/clipped_linear.py#L145

guyjacob:
You're correct, this is a bug indeed. When the input is 0 the output should be 1.
So, could be something like:
```python
output = torch.where(input == 0, torch.ones_like(input), torch.sign(input)) * E
```
If you'd like to try and test this and submit a quick PR that'll be great. Otherwise I'll try to get to it soon.

guyjacob:
Another option would be:
```python
output = input.sign() * E
output[output == 0] = 1
```
Not sure if there would be a performance difference between the two.

tk04m9:
Thank you for your quick response.  
Could you wait for a couple of days ?

PR #205, opened 2019-03-28
198 dorefa binarization

tacker-oh:
This PR is about #198 .
Fix `DorefaParamsBinarizationSTE` and Delete docstring about 1 bit special handling.

### DIFF
--- a/distiller/quantization/clipped_linear.py
+++ b/distiller/quantization/clipped_linear.py
@@ -142,8 +142,9 @@ class DorefaParamsBinarizationSTE(torch.autograd.Function):
         if inplace:
             ctx.mark_dirty(input)
         E = input.abs().mean()
-        output = input.sign() * E
+        output = input.sign()
         output[output == 0] = 1
+        output = output * E
         return output
     
     @staticmethod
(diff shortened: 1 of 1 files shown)


guyjacob:
Thanks for the PR!
Did you happen to check the other option I mentioned in the issue?
That is:
```python
output = torch.where(input == 0, torch.ones_like(input), torch.sign(input)) * E
```
I tried running preact-resnet18 training with 1-bit weights with the code as you pushed it (let's call it "indexing") and using `where`, and on my setup the `where` option was roughly 10% faster than indexing. Did you try / notice something else?

tacker-oh:
Sorry, I didn't try any comparison.
I compared two codes just now.
It takes 18m29s to train `resnet20_cifar` 200epochs with my code 
```
output = input.sign()
output[output == 0] = 1
output = output * E
```
and it takes 17m24s to train `resnet20_cifar` 200epochs with your option 
```
output = torch.where(input == 0, torch.ones_like(input), torch.sign(input)) * E
```
`torch.where()` is faster than my codes.

guyjacob:
Looks good, thanks for the PR @tacker-oh. Merging.

PR #207, opened 2019-03-28
Fixed typo

blairhan:


nzmora:
Hi @blairhan 
Thanks!
The source of the quantization documentation is stored here: https://github.com/NervanaSystems/distiller/blob/master/docs-src/docs/quantization.md
The HTML page that you fixed is actually produced by executing the instructions from here: 
https://github.com/NervanaSystems/distiller/blob/master/docs-src/README.md

After you do this, both the quantization.md and HTML file will be changed (perhaps a few other HTML files, due to new dates in some HTML files) and all of these changes need to be part of the PR.

We do it this way because the documentation is hosted on github pages that requires static pages from the git repo (these are the HTML pages).

Thanks!
Neta

blairhan:
@nzmora Hi! Thanks for the explanation.
I committed the changed the markdown page and generated HTML files as the instruction said.

PR #219, opened 2019-04-04
(BREAKING) save_checkpoint(): Save dict of user values instead of best_top1

guyjacob:
'best_top1' is too specific to classification workloads. Suggesting to just pass a dictionary of values instead, which will allow to pass anything the user deems worthy.
Also modified `load_checkpoint` to print a summary of the checkpoint contents, it looks something like this:

```
=> loading checkpoint logs/2019.04.04-131231/checkpoint.pth.tar
=> Checkpoint contents:
+----------------------+----------------+
| Key                  | Type / Value   |
|----------------------+----------------|
| arch                 | resnet20_cifar |
| best_epoch           | 1              |
| best_top1            | 58.06          |
| compression_sched    | dict           |
| current_top1         | 53.18          |
| epoch                | 2              |
| optimizer_state_dict | dict           |
| optimizer_type       | SGD            |
| state_dict           | OrderedDict    |
+----------------------+----------------+
```

For "simple" values it'll print the actual value, otherwise it prints the type.

### DIFF
--- a/distiller/apputils/checkpoint.py
+++ b/distiller/apputils/checkpoint.py
@@ -77,8 +77,7 @@ def save_checkpoint(epoch, arch, model, optimizer=None, scheduler=None,
     if hasattr(model, 'quantizer_metadata'):
         checkpoint['quantizer_metadata'] = model.quantizer_metadata
 
-    if extras:
-        checkpoint['extras'] = deepcopy(extras)
+    checkpoint['extras'] = deepcopy(extras)
 
     torch.save(checkpoint, fullpath)
     if is_best:
@@ -92,7 +91,7 @@ def load_lean_checkpoint(model, chkpt_file, model_device=None):
 
 def get_contents_table(d):
     def inspect_val(val):
-        if isinstance(val, (int, float, str)):
+        if isinstance(val, (Number, str)):
             return val
         elif isinstance(val, type):
             return val.__name__
--- a/distiller/apputils/checkpoint.py
+++ b/distiller/apputils/checkpoint.py
@@ -26,6 +26,7 @@ from errno import ENOENT
 import logging
 from numbers import Number
 from tabulate import tabulate
+from copy import deepcopy
 import torch
 import distiller
 from distiller.utils import normalize_module_name
@@ -33,7 +34,7 @@ msglogger = logging.getLogger()
 
 
 def save_checkpoint(epoch, arch, model, optimizer=None, scheduler=None,
-                    vals_to_save=None, is_best=False, name=None, dir='.'):
+                    extras=None, is_best=False, name=None, dir='.'):
     """Save a pytorch training checkpoint
 
     Args:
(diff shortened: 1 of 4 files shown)


guyjacob:
@barrh please take a look, can't add you as a reviewer for some reason

distiller/apputils/checkpoint.py
```diff
@@ -69,6 +75,8 @@ def save_checkpoint(epoch, arch, model, optimizer=None, scheduler=None,
     if hasattr(model, 'quantizer_metadata'):
         checkpoint['quantizer_metadata'] = model.quantizer_metadata
 
+    checkpoint.update(vals_to_save)
```
barrh: This will override checkpoint keys that were already created. e.g. optimizer_type
Is that what you intended for?

guyjacob: If a user passes one of the keywords we use then yeah, it'll override. Didn't think of that.
I guess there are 2 options:
1. Initialize checkpoint with the user passed values, then possibly override with ours (so loading doesn't break)
2. Check for "reserved" keywords and raise an error
Option 1 will just lead to confusion, so I'll add a check. We'll just have to remember to maintain this "reserved" list.

barrh: If you opt for option No. 2, I would opt for dynamic check on vals_to_save keys, right before the update():
if set(checkpoint) & set(vals_to_save): raise keys {} already exists

guyjacob: I thought of that but then we're allowing the user to use the keywords that are optional for us, that is anything other than 'epoch', 'arch' and 'state_dict'.
I think it's clearer if we're consistent about what's not allowed, and not having it change on a case-by-case basis.
Also, in case we don't use a specific keyword and then the user uses it with some unknown value, we'll try to do something with that unknown value at load time and the results of that are unexpected.
So static is the way to go, I'm afraid.

barrh: Perhaps just put vals_to_save in a different hierarchy: `checkpoint['user_appendix'] = vals_to_save`
It will resolve all possible conflicts.

guyjacob: That's good actually. Thanks!

distiller/apputils/checkpoint.py
```diff
@@ -26,14 +26,15 @@
 import logging
 from numbers import Number
```
nzmora: You can remove this import now.

distiller/apputils/checkpoint.py
```diff
@@ -42,18 +43,19 @@ def save_checkpoint(epoch, arch, model, optimizer=None, scheduler=None,
...
     """
     if not os.path.isdir(dir):
         raise IOError(ENOENT, 'Checkpoint directory does not exist at', os.path.abspath(dir))
 
-    if vals_to_save is None:
-        vals_to_save = {}
-    if not isinstance(vals_to_save, dict):
-        raise TypeError('vals_to_save must be either a dict or None')
+    if extras is None:
+        extras = {}
```
nzmora: why not keep it None?

distiller/apputils/checkpoint.py
guyjacob: Actually I should change the code the checks the type back to checking vs Number instead of explicit int and float. I copy-pasted `inspect_val` from the inspect_ckpt script and missed this little difference.

distiller/apputils/checkpoint.py
guyjacob: I think it's a remnant from other places where I used this "pattern", where I needed a valid dict even if the user didn't pass anything (using None as default since it's not a good idea to have mutable default arguments).

PR #225, opened 2019-04-08
Post-train quant: Extend activations clipping functionality

guyjacob:
* Some refactoring to enable multiple clipping methods
* BREAKING: clip_acts as a boolean flag (either in command line
  or in function signature) will fail. Error message with valid
  values from is displayed.
* Implemented clipping activations at mean + N * std
  (N is user configurable)
* Additional tests


distiller/quantization/q_utils.py
```diff
@@ -135,6 +135,22 @@ def get_tensor_avg_max_abs(t, across_dim=None):
     return torch.max(avg_min.abs_(), avg_max.abs_())
 
 
+def get_tensor_mean_n_stds_min_max(t, n_stds=1):
+    if n_stds <= 0:
+        raise ValueError('n_stds must be > 0, got {}'.format(n_stds))
+    mean = t.mean()
```
nzmora: if ```t``` has ```requires_grad=True``` then ```t.mean()``` will create an autograd subgraph (adding an edge from ```mean``` to ```t```). It's not a big deal here (the graph won't be traversed), just wasteful.  Maybe use ```t.data.mean()```?
We have the same happening in other places in the code (e.g. sparsity stats) that I plan to clean.

guyjacob: Hmm. You can say the same for nearly every function in q_utils. Maybe a decorator is a better solution than tracking down all the tensor operations. But let's discuss this separately. I need to think if there's any scenario where we we actually want these ops to be tracked by autograd (during a quant-aware training flow or something similar).

PR #227, opened 2019-04-14
Simplify deterministic evaluation

barrh:
Due to Pytorch's implementation of dataset loading, it is
_required_ to use single loading-worker in order to achieve
deterministic results. [1]
Since the user is not expected to be familiar with all the
quirks of deterministic execution, this patch lifts the
requirement for the user to set the number of workers to 1,
it does that automatically instead.

Another change that introduced here is that evaluation of
a model will be always deterministic.

Decreasing the number of workers **slows down** the evaluation.

Minor changes:
- utils.set_deterministic sets cudnn.benchmark as required [2]
- data_loaders.get_data_loaders enforces determinism correctly
- correct '--evaluate' help string

[1] - https://petewarden.com/2018/03/19/the-machine-learning-reproducibility-crisis/
[2] - https://pytorch.org/docs/stable/notes/randomness.html

### DIFF
--- a/distiller/apputils/data_loaders.py
+++ b/distiller/apputils/data_loaders.py
@@ -181,14 +181,6 @@ def get_data_loaders(datasets_fn, data_dir, batch_size, num_workers, validation_
         distiller.set_deterministic()
         worker_init_fn = __deterministic_worker_init_fn
 
-        # Experiment reproducibility is sometimes important.  Pete Warden expounded about this
-        # in his blog: https://petewarden.com/2018/03/19/the-machine-learning-reproducibility-crisis/
-        # In Pytorch, support for deterministic execution is still a bit clunky.
-        if num_workers > 1:
-            msglogger.warning('Number of data loader workers is decreased '
-                'to support deterministic execution')
-        num_workers = min(num_workers, 1)
-
     num_train = len(train_dataset)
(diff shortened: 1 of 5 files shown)


guyjacob:
I took this opportunity to check and it looks like the issue with multiple number of data loader workers 1 causing non-deterministic results was actually fixed in PyTorch long ago, in v0.4.0 (when we put this limitation in place in Distiller it was in v0.3.0 days). Shame on me/us for not checking this earlier...
I did a short test, running a few epochs of **training** of ResNet-20 on CIFAR using multiple workers (keeping all other determinism settings as-is) and got identical results across multiple runs.

So seems to me we can get rid of this num_workers == 1 limitation. And in that case, I have no issues with enforcing determinism when `--evaluate` is set.

Do you want to try this out for yourself to verify? And then if you don't mind, just re-purpose this PR to remove that logic completely?

(As a side-note - the Pete Warden blog post linked to in the comment in the code is about re-producibility in general, it has nothing to do PyTorch data-loading. So the reference as you made it above doesn't really make sense :). Also in the code, that comment should stay where it was in compress_classifier.py and not moved to data_loaders.py)

barrh:
@guyjacob feel free to edit the commit message

PR #236, opened 2019-04-29
Added PackedSequence functionality

levzlotnik:
* Added PackedSequence treatment in DistillerLSTM
* Added tests for the comparing torch.nn.LSTM treatment for PackedSequence


distiller/modules/rnn.py
```diff
@@ -231,7 +232,32 @@ def __init__(self, input_size, hidden_size, num_layers, bias=True, batch_first=F
     def forward(self, x, h=None):
         is_packed_seq = isinstance(x, nn.utils.rnn.PackedSequence)
         if is_packed_seq:
-            x, lengths = nn.utils.rnn.pad_packed_sequence(x, self.batch_first)
+            # Packed sequence treatment -
```
guyjacob: For clarity, suggest to extract a couple of functions - one for packed sequence handling and one for non-packed.

tests/test_lstm_impl.py
```diff
@@ -45,6 +46,10 @@ def test_conversion():
 def assert_output(out_true, out_pred):
     y_t, h_t = out_true
     y_p, h_p = out_pred
+    if isinstance(y_t, PackedSequence):
+        y_t, lenghts_t = pad_packed_sequence(y_t)
+        y_p, lenghts_p = pad_packed_sequence(y_p)
+        assert all(lenghts_t == lenghts_p)
     assert (y_t - y_p).abs().max().item() < ACCEPTABLE_ERROR
```
guyjacob: Not related to the specific change in the PR, but why not use `torch.testing.assert_allclose()`? See 'test_post_train_quant.py' for example. Use a more "standard" approach instead of re-inventing the wheel (as simple as that wheel may be).

PR #239, opened 2019-05-01
Fix shared module quantization

levzlotnik:
* Added check for shared module during `Quantizer._pre_process_container`
* In a case of a hit - log the hit and replace with the same wrapper as the previous reference


distiller/quantization/quantizer.py
```diff
@@ -173,6 +173,9 @@ def __init__(self, model, optimizer=None,
         self.train_with_fp_copy = train_with_fp_copy
         self.params_to_quantize = []
 
+        # A dictionary of replaced modules and their respective names.
+        self.modules_replaced = {}
```
guyjacob: Don't think there's a need to keep this around, is there? Just pass it as a parameter to `_pre_process_container`.

distiller/quantization/quantizer.py
```diff
@@ -226,6 +236,12 @@ def _pre_process_container(self, container, prefix=''):
         # Iterate through model, insert quantization functions as appropriate
         for name, module in container.named_children():
             full_name = prefix + name
+            if module in self.modules_replaced:
+                previous_name, previous_wrapper = self.modules_replaced[module]
+                msglogger.debug('Module {0} references to same module as {1},'
```
guyjacob: I actually think a warning is better here - we want the user to notice that something is being done that doesn't match the provided configuration.

distiller/quantization/quantizer.py
```diff
@@ -184,6 +187,13 @@ def _add_override_entry(self, module_name, entry):
         self.module_overrides_map[module_name] = entry
 
     def prepare_model(self):
+        """
+        Wrappes the model and all intended submodules with their quantized counterparts according to their
+        QBits and overrides configuration.
+        Note:
+            if a submodule is shared between different parts of the graph, but the configurations
+            in these parts are different - the first configuration is kept.
+        """
```
guyjacob: Bless you for adding documentation :)
It's not always wrapping, sometimes it's flat out replacing modules. Suggest the following:
```
"""
Traverses the model and replaces sub-modules with quantized counterparts according to the bit-width
and overrides configuration provided to __init__(), and according to the replacement_factory as
defined by the Quantizer sub-class being used.

Note:
    If multiple sub-modules within the model actually reference the same module, then that module
    is replaced only once, according to the configuration (bit-width and/or overrides) of the 
    first encountered reference.
    Toy Example - say a module is constructed using this bit of code:

        shared_relu = nn.ReLU
        self.relu1 = shared_relu
        self.relu2 = shared_relu
    
    When traversing the model, a replacement will be generated when 'self.relu1' is encountered.
    Let's call it `new_relu1'. When 'self.relu2' will be encountered, it'll simply be replaced
    with a reference to 'new_relu1'. Any override configuration made specifically for 'self.relu2' 
    will be ignored. A warning message will be shown.
"""
```
Turned out a bit long, but it's a confusing nuance so best to be clear about what what we're doing.

distiller/quantization/quantizer.py
levzlotnik: Keeping this around allows us more control and debug. Since failure won't happen (unlike what was initially discussed) - if something unexpected happens we can see more clearly where it went wrong.
Passing it as a parameter to the function is not that much different from passing it inside `self`, but after the function is done we don't need to return the dictionary (which wouldn't make sense anyways), and if we won't save this dict we won't have this info.
I'm thinking of changing it into an OrderedDict so we can be see better what was replaced when.

distiller/quantization/quantizer.py
levzlotnik: I agree.
However - in the Quantizer ctor we can clearly see that whatever the `model.named_modules()` doesn't see - the qbits/overrides don't even get recorded. So we can't know if the config between the different references is matching or not.

guyjacob: I agree, as it is it'll just be an indication to the user that something "weird" is happening.

distiller/quantization/quantizer.py
guyjacob: OrderedDict is a good idea.

But I'm still not convinced why we need to keep this around. Indeed we won't fail in case of different overrides for a shared module (I assume that's what you're referring to). But we do log a warning, so there is some indication that something happened. There's also the debug print in line 272 that logs all replacements made.
Then if you suspect something is wrong with the replacement process, then just debug that and inspect the history dict then and there.

PR #246, opened 2019-05-06
Refactor load/save checkpoint

barrh:
Add support for the following calls:
**save_checkpoint(model)** -> stores the model. 'arch' and 'dataset' are optional.
**load_checkpoint(file_path)** -> create the model. Return dict instead of fixed-size tuple. extra values included.
**load_checkpoint(file_path, model=create_model(...))** -> behaves like before. returns a dict.
**load_checkpoint(file_path, 'cpu')** -> load model to cpu

I tested this specific commit very lightly. passes pytest with Python3.6.
The 2nd commit in the PR only looks scary, but it only adds try-except.


nzmora:
closing due to staleness.

PR #251, opened 2019-05-13
Improved logging when saving collectors data

barrh:
Return file name from ActivationStatsCollector.save() and print it.

~~**Edit**: added pretrainedmodels Resnext101 support~~

### DIFF
--- a/examples/classifier_compression/compress_classifier.py
+++ b/examples/classifier_compression/compress_classifier.py
@@ -749,9 +749,8 @@ def create_activation_stats_collectors(model, *phases):
 
 def create_quantization_stats_collector(model):
     distiller.utils.assign_layer_fq_names(model)
-    return {'test': missingdict({'quantization_stats': QuantCalibrationStatsCollector(model, classes=None,
-                                                                                      inplace_runtime_check=True,
-                                                                                      disable_inplace_attrs=True)})}
+    return {'test': missingdict({'quantization_stats':
+                            QuantCalibrationStatsCollector(model, disable_inplace_attrs=True)})}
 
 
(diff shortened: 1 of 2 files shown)


examples/classifier_compression/compress_classifier.py
```diff
@@ -755,12 +755,11 @@ def create_quantization_stats_collector(model):
 
 
 def save_collectors_data(collectors, directory):
-    """Utility function that saves all activation statistics to Excel workbooks
+    """Utility function that saves all activation statistics to Excel workbooks or YAMLs
```
guyjacob:
```suggestion
    """Utility function that saves all activation statistics to disk.
    File type and format of contents are collector-specific.
```

examples/classifier_compression/compress_classifier.py
```diff
@@ -755,12 +755,11 @@ def create_quantization_stats_collector(model):
...
     """
     for name, collector in collectors.items():
-        workbook = os.path.join(directory, name)
-        msglogger.info("Generating {}".format(workbook))
-        collector.save(workbook)
+        workbook_path = collector.save(os.path.join(directory, name))
+        msglogger.info("Generating {}".format(workbook_path))
```
guyjacob: I think there was a reason to have the "Generating ..." line before the actual call to save() - for some collectors the save process could take a long time. So it's about letting the user know something is happening.
Perhaps something like:
```python
msglogger.info('Saving data for collector {}...'.format(name))
file_path = collector.save(os.path.join(directory, name))
msglogger.info('Saved to {}'.format(file_path))
```

PR #254, opened 2019-05-13
Activation Histograms

guyjacob:
Added a collector for activation histograms (sub-class of `ActivationStatsCollector`).
It is stats-based, meaning it requires pre-computed min/max stats per tensor. This is done in order to prevent the need to save all of the activation tensors throughout the run. The stats are expected in the format generated by `QuantCalibrationStatsCollector`.

Details:
* Implemented ActivationHistogramsCollector
* Added Jupyter notebook showcasing activation histograms
* Implemented helper function that performs the stats collection pass and histograms pass in one go
* Also added separate helper function just for quantization stats collection
* Integrated in image classification sample
* data_loaders.py: Added option to have a fixed subset throughout within the same session. Using it to keep the same subset between the stats collection and histograms collection phases.
* Other changes:
  * Calling assign_layer_fq_names in base-class of collectors. We do this since the collectors, as implemented so far, assume this is done. So makes sense to just do it in the base class instead of expecting the user to do it.
  * Enforcing a non-parallel model for quantization stats and histograms collectors
  * Jupyter notebook - add utility function to enable loggers in notebooks. This allows us to see any logging done by Distiller APIs called from notebooks.

### DIFF
--- a/distiller/apputils/data_loaders.py
+++ b/distiller/apputils/data_loaders.py
@@ -19,7 +19,6 @@
 This code will help with the image classification datasets: ImageNet and CIFAR10
 
 """
-import logging
 import os
 import torch
 import torchvision.transforms as transforms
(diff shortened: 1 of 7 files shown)


guyjacob:
@barrh:
1. When #251 is merged I'll merge it to this branch and add it to the new collector as well. Can you take a look at the comments I left you there?
2. The purpose here is to have something that can be used "offline" for other purposes. For example, there are methods to determine activation clipping values based on histograms, which we can now add. Visualizing the histogram is another purpose. Images are just a straight forward way to do that. Admittedly I didn't consider doing it with TensorBoard, but I think that's more geared towards tracking the histograms through training, which is not what I'm focused on at the moment. I'm just looking for a "static" snapshot.

PR #258, opened 2019-05-14
Rename onnx

barrh:
I also applied some cleanup on related code. So there're logical changes in this PR.
I didn't try to import the exported models.

### DIFF
--- a/distiller/model_summaries.py
+++ b/distiller/model_summaries.py
@@ -52,8 +52,8 @@ def model_summary(model, what, dataset=None):
     elif what == 'compute':
         try:
             dummy_input = dataset_dummy_input(dataset)
-        except ValueError:
-            print("Unsupported dataset (%s) - aborting compute operation" % dataset)
+        except ValueError as e:
+            print(e)
             return
         df = model_performance_summary(model, dummy_input, 1)
         t = tabulate(df, headers='keys', tablefmt='psql', floatfmt=".5f")
@@ -431,16 +431,19 @@ def draw_img_classifier_to_file(model, png_fname, dataset, display_param_nodes=F
                                    'fillcolor': 'gray',
                                    'style': 'rounded, filled'}
     """
+    dummy_input = dataset_dummy_input(dataset)
     try:
-        dummy_input = dataset_dummy_input(dataset)
-        model = distiller.make_non_parallel_copy(model)
-        g = SummaryGraph(model, dummy_input)
+        non_para_model = distiller.make_non_parallel_copy(model)
+        g = SummaryGraph(non_para_model, dummy_input)
+
         draw_model_to_file(g, png_fname, display_param_nodes, rankdir, styles)
         print("Network PNG image generation completed")
     except FileNotFoundError:
         print("An error has occured while generating the network PNG image.")
         print("Please check that you have graphviz installed.")
         print("\t$ sudo apt-get install graphviz")
+    finally:
+        del non_para_model
 
 
 def dataset_dummy_input(dataset):
@@ -449,7 +452,7 @@ def dataset_dummy_input(dataset):
     elif dataset == 'cifar10':
         dummy_input = torch.randn(1, 3, 32, 32)
     else:
-        raise ValueError("Unsupported dataset (%s) - aborting draw operation" % dataset)
+        raise ValueError("Unsupported dataset (%s) - aborting operation" % dataset)
     return dummy_input
 
 
--- a/distiller/model_summaries.py
+++ b/distiller/model_summaries.py
@@ -27,7 +27,6 @@ import pandas as pd
 from tabulate import tabulate
 import logging
 import torch
-from torch.autograd import Variable
 import torch.optim
 import distiller
 from .summary_graph import SummaryGraph
@@ -49,11 +48,9 @@ def model_summary(model, what, dataset=None):
         csvlogger = CsvLogger()
         distiller.log_weights_sparsity(model, -1, loggers=[pylogger, csvlogger])
     elif what == 'compute':
-        if dataset == 'imagenet':
-            dummy_input = Variable(torch.randn(1, 3, 224, 224))
-        elif dataset == 'cifar10':
-            dummy_input = Variable(torch.randn(1, 3, 32, 32))
-        else:
+        try:
+            dummy_input = dataset_dummy_input(dataset)
+        except ValueError:
             print("Unsupported dataset (%s) - aborting compute operation" % dataset)
             return
         df = model_performance_summary(model, dummy_input, 1)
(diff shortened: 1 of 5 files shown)


examples/classifier_compression/parser.py
```diff
@@ -77,9 +77,10 @@ def get_parser():
                         help='print masks sparsity table at end of each epoch')
     parser.add_argument('--param-hist', dest='log_params_histograms', action='store_true', default=False,
                         help='log the parameter tensors histograms to file (WARNING: this can use significant disk space)')
-    parser.add_argument('--summary', type=lambda s: s.lower(), choices=SUMMARY_CHOICES,
-                        help='print a summary of the model, and exit - options: ' +
-                        ' | '.join(SUMMARY_CHOICES))
+    parser.add_argument('--summary', type=lambda s: s.lower(), choices=SUMMARY_CHOICES, action='append',
+                        help='print a summary of the model, and exit - options: | '.join(SUMMARY_CHOICES))
+    parser.add_argument('--export-onnx', action='store', nargs='?', type=str, const='model.onnx', default=None,
```
nzmora: why is nargs=?

distiller/model_summaries.py
```diff
@@ -446,30 +443,27 @@ def draw_img_classifier_to_file(model, png_fname, dataset, display_param_nodes=F
...
     elif dataset == 'cifar10':
-        dummy_input = Variable(torch.randn(1, 3, 32, 32))
+        dummy_input = torch.randn(1, 3, 32, 32)
     else:
         raise ValueError("Unsupported dataset (%s) - aborting draw operation" % dataset)
     return dummy_input
 
 
-def export_img_classifier_to_onnx(model, onnx_fname, dataset, export_params=True, add_softmax=True):
+def export_img_classifier_to_onnx(model, onnx_fname, dataset, **kwargs):
```
nzmora: You removed the option to add a softmax layer.  What's the reason?

examples/classifier_compression/parser.py
barrh: That's a trick: I wanted to allow calling with or without the explicit filename. e.g. Calling `--export-onnx` should act like: `--export-onnx model.onnx`, both are valid options.
(for more information: refer https://docs.python.org/3/library/argparse.html#nargs)
The only bug I found for this approach is when `--export-onnx` precedes the dataset path. e.g. `python classifier.py --export-onnx IMAGENET_PATH` will fail unexpectedly.

distiller/model_summaries.py
barrh: 1. I went back and explored the reason for you adding this option, and I figured that if Alexnet (and I have tested couple more topologies too) is exported without errors, then we're good.
2. I intentionally removed code that is irrelevant in Pytorch v1.0.1 (or v1.1), is that a bad thing?

nzmora: I added the explicit handling of softmax because TorchVision models use nn.CrossEntropyLoss for computing the loss, instead of adding a softmax layer and computing log-softmax.  
When we export to ONNX to run on HW, we often want to have the softmax layer added to the "end" of the model.  I don't think this has changed between pytorch versions, so it is still required logic.

barrh:
I couldn't test the final commit (b12298c) due to tk requirement, which I cannot fulfill.

distiller/model_summaries.py
```diff
@@ -446,30 +445,37 @@ def draw_img_classifier_to_file(model, png_fname, dataset, display_param_nodes=F
...
+        kwargs: arguments to be passed to torch.onnx.export
     """
     dummy_input = dataset_dummy_input(dataset).to('cuda')
-    # Pytorch 0.4 doesn't support exporting modules wrapped in DataParallel
-    model = distiller.make_non_parallel_copy(model)
+    # Pytorch doesn't support exporting modules wrapped in DataParallel
+    non_para_model = distiller.make_non_parallel_copy(model)
 
-    with torch.onnx.set_training(model, False):
+    try:
```
nzmora: why did you remove setting ONNX to inference mode?

distiller/model_summaries.py
```diff
@@ -40,20 +39,20 @@
...
         distiller.log_weights_sparsity(model, -1, loggers=[pylogger, csvlogger])
     elif what == 'compute':
-        if dataset == 'imagenet':
-            dummy_input = Variable(torch.randn(1, 3, 224, 224))
-        elif dataset == 'cifar10':
-            dummy_input = Variable(torch.randn(1, 3, 32, 32))
-        else:
+        try:
+            dummy_input = dataset_dummy_input(dataset)
+        except ValueError:
```
nzmora: We are throwing an exception, so why catch it only to print a message?  I would remove the try except and change the ValueError message

distiller/model_summaries.py
```diff
@@ -446,30 +445,37 @@ def draw_img_classifier_to_file(model, png_fname, dataset, display_param_nodes=F
 
 def dataset_dummy_input(dataset):
     if dataset == 'imagenet':
-        dummy_input = Variable(torch.randn(1, 3, 224, 224), requires_grad=False)
+        dummy_input = torch.randn(1, 3, 224, 224)
     elif dataset == 'cifar10':
-        dummy_input = Variable(torch.randn(1, 3, 32, 32))
+        dummy_input = torch.randn(1, 3, 32, 32)
     else:
         raise ValueError("Unsupported dataset (%s) - aborting draw operation" % dataset)
```
nzmora:
```suggestion
        raise ValueError("Unsupported dataset (%s) - aborting operation" % dataset)
```

distiller/model_summaries.py
```diff
@@ -446,30 +445,37 @@ def draw_img_classifier_to_file(model, png_fname, dataset, display_param_nodes=F
...
     return dummy_input
 
 
-def export_img_classifier_to_onnx(model, onnx_fname, dataset, export_params=True, add_softmax=True):
+def export_img_classifier_to_onnx(model, onnx_fname, dataset, add_softmax=True, **kwargs):
```
nzmora: You removed 'verbose=False, export_params=export_params'.
This gives the caller more power, but also burdens them with more details.  
Do the ONNX kwargs defaults match these values?

If the defaults do not match, I would prefer that you have one function with kwargs, and one with 'verbose=False, export_params=export_params' that the application calls.  This also makes sure that the application will be protected from a pytorch upgrade (if an upgrade changes the values of the defaults)

examples/classifier_compression/compress_classifier.py
```diff
@@ -196,7 +196,14 @@ def main():
 
     # This sample application can be invoked to produce various summary reports.
     if args.summary:
-        return summarize_model(model, args.dataset, which_summary=args.summary)
+        for s in which_summary:
+            distiller.model_summary(model, s, dataset)
+        return
+
+    if args.export_onnx is not None:
```
nzmora: when --export-onnx is used from the command line - w/o an onnx model name parameter, what is the value of args.export_onnx?  Is it an empty string?

distiller/model_summaries.py
barrh: I didn't want to change the current behavior. But this could be done.

examples/classifier_compression/compress_classifier.py
barrh: no, it defaults to 'model.onnx'

distiller/model_summaries.py
barrh: Same reasoning as with the softmax... It passes my tests without it.

distiller/model_summaries.py
barrh: In pytorch v1.0.1 and in master, the defaults match.

distiller/model_summaries.py
barrh: I think the indentation is wrong. I can fix it in another last commit.

distiller/model_summaries.py
nzmora: It's better to add it back - see the documentation: "A context manager to temporarily set the training mode of 'model' to 'mode', resetting it when we exit the with-block.  A no-op is mode is None."
https://github.com/pytorch/pytorch/blob/master/torch/onnx/utils.py

Your test doesn't test for the correctness of the export, so it would not fail.  Also, it is likely that the model you're testing is in "inference" mode

nzmora: [This](https://github.com/pytorch/pytorch/blob/master/torch/onnx/utils.py) documentation of onnx.export says:
"training (bool, default False): export the model in training mode.  At
            the moment, ONNX is oriented towards exporting models for inference
            only, so you will generally not need to set this to True."
So onnx.set_training is not currently needed.
In fact, I don't think it is needed at all if we explicitly set 'training=False' when invoking onnx.export.

PR #262, opened 2019-05-19
Store qe_stats_file in output directory

barrh:
Stores the qe_stats_file, under configs.
Extend log_execution_env_state to accept
list of configuration files.

### DIFF
--- a/distiller/apputils/execution_env.py
+++ b/distiller/apputils/execution_env.py
@@ -20,7 +20,6 @@ This is helpful if you want to recreate an experiment at a later time, or if
 you want to understand the environment in which you execute the training.
 """
 
-import contextlib
 import logging
 import logging.config
 import os
@@ -41,16 +40,16 @@ except ImportError:
 logger = logging.getLogger("app_cfg")
 
 
-def log_execution_env_state(config_path=None, logdir=None, gitroot='.'):
+def log_execution_env_state(config_paths=None, logdir=None, gitroot='.'):
     """Log information about the execution environment.
 
-    File 'config_path' will be copied to directory 'logdir'. A common use-case
+    Files in 'config_paths' will be copied to directory 'logdir'. A common use-case
     is passing the path to a (compression) schedule YAML file. Storing a copy
     of the schedule file, with the experiment logs, is useful in order to
     reproduce experiments.
 
     Args:
-        config_path: path to config file, used only when logdir is set
+        config_paths: path(s) to config file(s), used only when logdir is set
         logdir: log directory
         git_root: the path to the .git root directory
     """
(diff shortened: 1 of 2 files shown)


distiller/apputils/execution_env.py
```diff
@@ -91,19 +91,22 @@ def log_git_state():
 
     if (logdir is None) or (config_path is None):
         return
+    if isinstance(config_path, str) or not hasattr(config_path, '__iter__'):
```
nzmora: I think that we can narrow the interface and say you can pass either a string or a list of strings, but I'm not going to insist on this point.
However, I prefer that you wrap this logic in a function (e.g. is_iterable) to make the main logic an easy read:
```
def is_iterable(x):
    return hasattr(obj, '__iter__') and not isinstance(obj, str)
.
.
if not is_iterable(config_path):
    config_path = list(config_path)
```

distiller/apputils/execution_env.py
```diff
@@ -91,19 +91,22 @@ def log_git_state():
...
-    if os.path.exists(os.path.join(configs_dest,
-                                   os.path.basename(config_path))):
-        logger.debug('{} already exists in logdir'.format(
-            os.path.basename(config_path) or config_path))
-    else:
-        try:
-            shutil.copy(config_path, configs_dest)
-        except OSError as e:
-            logger.debug('Failed to copy of config file: {}'.format(str(e)))
+    for conf in config_path:
```
nzmora: I think that now you can remove the line that imports contextlib

distiller/apputils/execution_env.py
```diff
@@ -91,19 +91,22 @@ def log_git_state():
 
     if (logdir is None) or (config_path is None):
         return
+    if isinstance(config_path, str) or not hasattr(config_path, '__iter__'):
+        config_path = [config_path]
```
nzmora: Since you are creating a collection (list), it would be clearer (especially in the code below) if you "pluralized" the variable name (i.e. ```config_paths``` instead of ```config_path```).

distiller/apputils/execution_env.py
nzmora: it's clearer to use ```path``` or ```conf_path``` or ```config_path``` or ```cpath``` - over ```conf``` which obscures the fact that this is a string that holds a path to a file.

distiller/apputils/execution_env.py
barrh: Strings are also iterable, this is why I had to exclude it in the condition. Forming a function named is_iterable for this condition would be misleading.
Also iterable (at least in Py3.6): generators and alike.
`list()` and [] operator have different behaviors, I think we should stick with `[]`

guyjacob: If readability is the concern, I find the following a bit clearer (from [here](https://stackoverflow.com/a/1952481)):
```python
from collections.abc import Iterable

if isinstance(config_path, str) or not isinstance(config_path, Iterable):
```
To me that's good enough. I'm always in favor of extracting functions, but I agree that `is_iterable` is not a good name in this case, and calling it `is_iterable_but_not_str` seems a bit too much :)

distiller/apputils/execution_env.py
guyjacob: I second this

PR #269, opened 2019-05-26
Added support for setting the PRNG seed

nzmora:
Added set_seed() to Distiller (utils.py) and added support for seeding the PRNG,
in both reproducible and non-reproducible modes.
The PRNGs of Pytorch (CPU & Cuda devices), numpy and Python are set.
Added support for ```--seed``` to classifier_compression.py.

### DIFF
--- a/distiller/utils.py
+++ b/distiller/utils.py
@@ -595,18 +595,35 @@ def make_non_parallel_copy(model):
     return new_model
 
 
-def set_deterministic():
-    msglogger.debug('set_deterministic is called')
-    torch.manual_seed(0)
-    random.seed(0)
-    np.random.seed(0)
-    torch.backends.cudnn.deterministic = True
-    torch.backends.cudnn.benchmark = False
+def set_seed(seed):
+    """Seed the PRNG for the CPU, Cuda, numpy and Python"""
+    torch.manual_seed(seed)
+    random.seed(seed)
+    np.random.seed(seed)
+
+
+def set_deterministic(is_deterministic=True, seed=0):
+    '''Try to configure the system for reproducible results.
+
+    Experiment reproducibility is sometimes important.  Pete Warden expounded about this
+    in his blog: https://petewarden.com/2018/03/19/the-machine-learning-reproducibility-crisis/
+    For Pytorch specifics see: https://pytorch.org/docs/stable/notes/randomness.html#reproducibility
+    '''
+    msglogger.debug("set_deterministic({})".format(is_deterministic))
+    if seed is None and is_deterministic:
+        seed = 0
+    if seed is not None:
+        set_seed(seed)
+    torch.backends.cudnn.deterministic = is_deterministic
+    # Turn on CUDNN benchmark mode for best performance (non-deterministic). This is usually "safe"
+    # for image classification models, as the input sizes don't change during the run
+    # See here: https://discuss.pytorch.org/t/what-does-torch-backends-cudnn-benchmark-do/5936/3
+    torch.backends.cudnn.benchmark = not is_deterministic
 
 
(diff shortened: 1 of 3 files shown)


distiller/utils.py
```diff
@@ -595,18 +595,35 @@ def replace_data_parallel(container):
...
+    msglogger.debug("set_deterministic({})".format(is_deterministic))
+    if seed is None and is_deterministic:
+        seed = 0
+    if seed is not None:
+        set_seed(seed)
+    torch.backends.cudnn.deterministic = is_deterministic
+    # Turn on CUDNN benchmark mode for best performance (non-deterministic). This is usually "safe"
+    # for image classification models, as the input sizes don't change during the run
+    # See here: https://discuss.pytorch.org/t/what-does-torch-backends-cudnn-benchmark-do/5936/3
+    torch.backends.cudnn.benchmark = not is_deterministic
```
guyjacob: The link in the comment explains that benchmark mode is not desirable in cases where input size varies between iterations. That could be language/speech models for example. That's the reason I left it out of this function originally. Left it to the user.
Was your intention to just toggle it back to what it was before? If so then maybe we could cache it.

nzmora: My intention was to set ```torch.backends.cudnn.benchmark = not is_deterministic```.
From your comment I understand this is half true - i.e. if ```is_deterministic == True``` ==> ```torch.backends.cudnn.benchmark = False```, but if ```is_deterministic == False``` ==> ```torch.backends.cudnn.benchmark = ``` depends on the application.
So it makes sense to revert my logic.  Argh ;-)

guyjacob: It does leave the burden with the user, and I'm guessing many users aren't even aware of this. But still...

lehahoang:
Hi guys,

Thank you for adding this amazing feature. 
However, I have found out that the seed is still set to zero during the evaluation mode. I guess the root cause is we set it twice in `compress_classifier.py`.
Fortunately, by moving the code section in `compress_classifier.py` which sets the PRNG seed (from line 106 to 116 in original script) after this line ` train_loader, val_loader, test_loader, _ = load_data(args)` (line 222 in original script ), the seed now is set by our args.seed.
Here is how I put them together in the modified script:

```
    if args.qe_calibration:
        return acts_quant_stats_collection(model, criterion, pylogger, args)

    if args.activation_histograms:
        return acts_histogram_collection(model, criterion, pylogger, args)  
    train_loader, val_loader, test_loader, _ = load_data(args)
    msglogger.info('Dataset sizes:\n\ttraining=%d\n\tvalidation=%d\n\ttest=%d',
                   len(train_loader.sampler), len(val_loader.sampler), len(test_loader.sampler))
    if args.evaluate:
        args.deterministic = True
    if args.deterministic:
        distiller.set_deterministic(args.seed) # For experiment reproducability
    else:
        if args.seed is not None:
            distiller.set_seed(args.seed)
        # Turn on CUDNN benchmark mode for best performance. This is usually "safe" for image
        # classification models, as the input sizes don't change during the run
        # See here: https://discuss.pytorch.org/t/what-does-torch-backends-cudnn-benchmark-do/5936/3
        cudnn.benchmark = True
    if args.sensitivity is not None:
        sensitivities = np.arange(args.sensitivity_range[0], args.sensitivity_range[1], args.sensitivity_range[2])
        return sensitivity_analysis(model, criterion, test_loader, pylogger, args, sensitivities)

    if args.evaluate:
        return evaluate_model(model, criterion, test_loader, pylogger, activations_collectors, args,
                              compression_scheduler)
```

It might be a dummy solution but I did not come up with a better one. Please let me know if you have a simpler solution just like using the arguments. Thank you.

One more thing I am still confused is the usage of the argument `--deterministic`. Is it for the deterministic training process?
Please correct my assumption if I am wrong. 
Best regards,
Le Ha

PR #292, opened 2019-06-18
Input shape attribute in models

guyjacob:
Setting this attribute will become a requirement for some upcoming features

* Add utility function to set input_shape attribute in models
* Set this attribute in the classification models factory `distiller.models.create_model()`
* Add shape parameter in get_dummy_input (in addition to dataset)

### DIFF
--- a/distiller/apputils/data_loaders.py
+++ b/distiller/apputils/data_loaders.py
@@ -49,11 +49,11 @@ def classification_num_classes(dataset):
 
 def classification_get_input_shape(dataset):
     if dataset == 'imagenet':
-        return [1, 3, 224, 224]
+        return 1, 3, 224, 224
     elif dataset == 'cifar10':
-        return [1, 3, 32, 32]
+        return 1, 3, 32, 32
     elif dataset == 'mnist':
-        return [1, 1, 28, 28]
+        return 1, 1, 28, 28
     else:
         raise ValueError("dataset %s is not supported" % dataset)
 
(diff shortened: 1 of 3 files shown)


distiller/utils.py
```diff
@@ -557,8 +557,47 @@ def has_children(module):
...
+    return dummy_input
+
+
+def set_model_input_shape_attr(model, dataset=None, input_shape=None):
+    """Sets an attribute named 'input_shape' within the model instance, specifying the expected input shape
+
+    Args:
+          model (nn.Module): Model instance
+          dataset (str): Name of dataset from which to infer input shape
+          input_shape (list): List of integers representing the input shape. Used only if 'dataset' is None
```
nzmora: consider using tuples instead of lists.  It's just a bit cleaner IMHO, since they are read-only.

distiller/apputils/data_loaders.py
```diff
@@ -47,22 +47,15 @@ def classification_num_classes(dataset):
             'imagenet': 1000}.get(dataset, None)
 
 
-def classification_get_dummy_input(dataset, device=None):
-    """Generate a representative dummy (random) input for the specified dataset.
-
-    If a device is specified, then the dummay_input is moved to that device.
-    """
+def classification_get_input_shape(dataset):
```
nzmora: I was too lazy to move classification_get_dummy_input to utils, but I think you should do it now ;-).
Or maybe it's time to finally refactor utils.py.

distiller/utils.py
guyjacob: Done

PR #300, opened 2019-06-24
Allow loading checkpoints with superfluous state-keys

nzmora:
Change the default checkpoint loading behavior, such that we allow
loading a checkpoint that has state-dict entries that do not
appear in the original (Distiller) model.

### DIFF
--- a/distiller/apputils/checkpoint.py
+++ b/distiller/apputils/checkpoint.py
@@ -101,7 +101,8 @@ def get_contents_table(d):
     return tabulate(contents, headers=["Key", "Type", "Value"], tablefmt="psql")
 
 
-def load_checkpoint(model, chkpt_file, optimizer=None, model_device=None, *, lean_checkpoint=False):
+def load_checkpoint(model, chkpt_file, optimizer=None, model_device=None, *, 
+                    lean_checkpoint=False, ignore_superfluous_loaded_keys=True):
     """Load a pytorch training checkpoint.
 
     Args:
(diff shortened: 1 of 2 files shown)


distiller/apputils/checkpoint.py
```diff
@@ -164,7 +165,13 @@ def load_checkpoint(model, chkpt_file, optimizer=None, model_device=None, *, lea
 
     if normalize_dataparallel_keys:
             checkpoint['state_dict'] = {normalize_module_name(k): v for k, v in checkpoint['state_dict'].items()}
-    model.load_state_dict(checkpoint['state_dict'])
+
+    loaded_dict = checkpoint['state_dict']
```
guyjacob: Should be `distiller_model_dict`? Otherwise if ignore is False then line 173 will fail.

distiller/apputils/checkpoint.py
```diff
@@ -164,7 +165,13 @@ def load_checkpoint(model, chkpt_file, optimizer=None, model_device=None, *, lea
...
+    if ignore_superfluous_loaded_keys:
+        # filter out unnecessary keys
+        distiller_model_dict = {k: v for k, v in loaded_dict.items() if k in model.state_dict().keys()}
+    model.load_state_dict(distiller_model_dict)
```
guyjacob: There's the `strict` flag in [load_state_dict](https://pytorch.org/docs/stable/nn.html#torch.nn.Module.load_state_dict), I think you need to set it to False if you want this to work.
Actually - If you use the `strict` flag you don't even need to do the filtering you do above.

guyjacob: Disclaimer - never used this `strict` flag but looking at the docs I think it does what you're looking for

nzmora: You are correct on all counts.

PR #303, opened 2019-07-02
Bugfix: convert thresholds to weights' dtype

sgrechanik-h:
This fixes `RuntimeError: Expected object of scalar type Half but got scalar type Float for argument #2 'other'`, e.g. during fp16 training.


dvolgyes:
It seems the code changed quite a bit in the meantime.
It would be nice to run some fp16 tests, and if they work, then close this PR.

nzmora:
I'm closing this PR because indeed it has been too long and the code has changed quite a bit.  As @dvolgyes suggested, for a full PR we would need to run tests using fp16 and at present we are not planning to do this.

Issue #240, opened 2019-05-01
support ReduceLRonPlateau(mode='max') in classifier.py


barrh:
Currently, during training, only loss value is passed to the LR scheduler, but if the scheduler is ReduceLRonPlateau with mode='max', it doesn't make sense to pass loss. Top1 should be passed instead.

barrh:
merged.

PR #306, opened 2019-07-04
Switch to PyTorch 1.1.0

guyjacob:
* PyTorch 1.1.0 now required, moved other dependencies to up-to-date versions as well
* Adapt LR scheduler to PyTorch 1.1 API changes:
  - Change lr_scheduler.step() calls to succeed validate calls, during training
  - Pass to lr_scheduler.step() caller both loss and top1 (Resolves issue #240)
* Adapt thinning for PyTorch 1.1 semantic changes
  - **KNOWN ISSUE**: When a thinning recipe is applied, in certain cases PyTorch displays this warning: "UserWarning: non-inplace resize is deprecated". To be fixed later.
* SummaryGraph: Workaround for new scope name issue from PyTorch 1.1.0
* Adapt to updated PyTest version:
  - Stop using deprecated 'message' parameter of pytest.raises(), use pytest.fail() instead
  - Make sure only a single test case per pytest.raises context
* Move PyTorch version check to root \_\_init\_\_.py 
  - This means the version each checked when Distiller is first imported. A RuntimeError is raised if the version is wrong.
* Updates to parameter_histograms notebook:
  - Replace deprecated normed argument with density
  - Add sparsity rate to plot title
  - Load model in CPU


guyjacob:
Reviewed offline, good to go

PR #307, opened 2019-07-04
Allow loading checkpoints with superfluous state-keys

nzmora:
Change the default checkpoint loading behavior, such that we allow
loading a checkpoint that has state-dict entries that do not
appear in the original (Distiller) model.


guyjacob:
@nzmora - Didn't we already merge this in #300? When I merge master into this branch, it shows no diffs. Keeping open at the moment in case I'm missing something...

nzmora:
@guyjacob ,

Yup - we did merge this.  I don't know what happened here, but i'm closing.

PR #320, opened 2019-07-15
Increase Tensorflow min version to 1.14

barrh:
tensorboard 1.14 is required to take advantage of Pytorch's exp. implementation of summarywriter.
(It wasn't stable yet when I pushed the Pytorch 1.1 initial commit)


guyjacob:
LGTM. Are you trying to replace the Distiller TensorBoard logger with the PyTorch implementation?

barrh:
@guyjacob Not immediately, I need to try it out first. In theory, it seems like the right way to go hence forth.

PR #321, opened 2019-07-16
Support quantize weights/bias only

hunterkun:
I add support quantizing weights only in the post training quantization when a user set `bits_activations=32`. It directly uses quantize-dequantize operations on weights and keep activations fp32 and no warp on layer. I use literally 32 instead of None because it would conflict with this snippet https://github.com/NervanaSystems/distiller/blob/master/distiller/quantization/quantizer.py#L277
So now, i can't think of a better way to denote thought it may looks weird.

### DIFF
--- a/distiller/quantization/quantizer.py
+++ b/distiller/quantization/quantizer.py
@@ -274,6 +274,11 @@ class Quantizer(object):
                     replace_msg(full_name)
                 continue
             current_qbits = self.module_qbits_map[full_name]
+            if current_qbits.acts == 32 and current_qbits.wts:
+                valid_kwargs, invalid_kwargs = distiller.filter_kwargs(self.module_overrides_map[full_name], self.quantize_weights)
+                self.quantize_weights(module, current_qbits.wts, **valid_kwargs)
+                self.modules_processed[module] = full_name, None
+                continue
             if current_qbits.acts is None and current_qbits.wts is None:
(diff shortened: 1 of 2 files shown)


hunterkun:
Yeah, thanks for your good advice. But the conflict is still there.
https://github.com/NervanaSystems/distiller/blob/master/distiller/quantization/quantizer.py#L278
Sorry for not elaborating on this. Now I explain it why.
`for name, module in container.named_children():`, In this for loop, the first time coming into loop , `container` equals whole Model, and it would never match this `if current_qbits.acts is None and current_qbits.wts is None:` and so it can go on recursively. It is OK. But things change when you would like to set activation bits to None. It would match that sentence since this sentence https://github.com/NervanaSystems/distiller/blob/master/distiller/quantization/quantizer.py#L187 and then it early exits this loop **just one time**, which could not replace module.
Now, I could not think of a better way except change this logic.

guyjacob:
Hi @hunterkun, sorry again for the delayed response.
You're right, we weren't correctly handling the case where the user doesn't want to quantize activations by default. I pushed a PR to fix that - #345. While it's being reviewed, can you try merging it to your code and continue from there? I think that with this fix my initial comments will be valid again.

hunterkun:
Sorry for deleting the old repo by mistake since i'm not familiar with git operation. I have reopened new pr.

PR #325, opened 2019-07-18
fix thinning fc after resblock

bezorro:
This bug is triggered when you perform Filter Pruning of the ResBlock just before the FC in [ResNet50](https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py).

If you prune [layer4.0.conv3.weight, layer4.1.conv3.weight, layer4.2.conv3.weight, layer4.0.downsample] at the same time(layer4.0.conv3.weight is the Leader), the successor FC would be thinned several times. This code `if param.data.size(dim) != len_indices:` prevents the occurrence. But `selection_view = param.view(*directive[2])` is placed before the IF Statement, which cause the bug: when you thin `param` at the second time, the size of `param` and `directive[2]` does not match so the bug occurs.

Thus, I simply move `selection_view = param.view(*directive[2])` after `if param.data.size(dim) != len_indices:`.


nzmora:
Looks good - thanks!
Can you add a test that fails without the fix and succeeds with the new code (to [here](https://github.com/NervanaSystems/distiller/blob/master/tests/test_pruning.py))?
I'd like to have a regression test so that we don't repeat this bug in the future.

Thanks!
Neta

PR #328, opened 2019-07-21
Workaround for non 1:1 mapping between model and SummaryGraph when nn.ModuleList is used

guyjacob:
The PyTorch trace mechanism doesn't "see" `torch.nn.ModuleList` modules (since they don't have a `forward` function). As a result, the mapping from module names at the Python model definition level to the scope-names at the trace level is not 1:1. This makes it impossible for us to map back from `SummaryGraph` ops to their respective `nn.Module`s, which is required for flows like BatchNorm folding and stats fusion in post-training quantization.

In #313 we handled this issue specifically in `DistillerLSTM`, but it makes much more sense to have a generic and automatic solution for this issue, which doesn't require the user to modify the model. This is such a solution.

Details:
* Implemented `DistillerModuleList`, a replacement for `torch.nn.ModuleList` which results in full and unique scope-names
* See documentation for this class in `summary_graph.py` for extensive details on the issue and solution
* When generating a `SummaryGraph`, the model is scanned and all instances of `torch.nn.ModuleList` are replaced with `DistillerModulelist`
* Add tests for new functionality
* Revert `DistillerLSTM` to use the standard `torch.nn.ModuleList`, as now we're handling this issue automatically, and no need to confuse users with ad-hoc list implementations


distiller/modules/rnn.py
```diff
@@ -201,45 +201,34 @@ def __init__(self, input_size, hidden_size, num_layers, bias=True, batch_first=F
                 # # Process each timestep at the entire layers chain -
                 # # each timestep is forwarded through `front` and `back` chains independently,
                 # # similarily to a unidirectional LSTM.
-                # self.cells = self._create_cells_list('cell', 1)
-                # self.cells_reverse = self._create_cells_list('cell_reverse', 2)
+                # self.cells = self._create_cells_list(1)
+                # self.cells_reverse = self._create_cells_list(2)
```
nzmora: Do we need to keep this function (`_layer_chain_bidirectional_type1`) with all of the code remarked?

distiller/summary_graph.py
```diff
@@ -561,3 +572,169 @@ def __eq__(self, other):
...
+      * This means that in terms of accessing the modules and invoking them, DistillerModuleList behaves exactly the
+        same as torch.nn.ModuleList. See the example below.
+
+    Differences vs. torch.nn.ModuleList:
+      * DistillerModuleList is NOT a sub-class of torch.nn.Module
+      * This means that the modules in the list are NOT sub-modules of the list itself. They are registered as
+        sub-modules of the "parent module". That is - the contents of a DistillerModuleList are "flattened" within the
+        "parent module".
+      * In addition, we can't use the '.' character to denote the "nesting" of a module within the list. We use '_'.
+      * All of This means that calls to functions like state_dict() / named_modules() / named_children() / etc. on the
```
nzmora: "All of This" ==> "All of this"

distiller/summary_graph.py
```diff
@@ -561,3 +572,169 @@ def __eq__(self, other):
...
+        "parent_module" return different results when this class is used compared to torch.nn.ModuleList.
+
+    At the moment we don't see a usage for this class "in the wild", outside of SummaryGraph generation.
+    In the context of SummaryGraph, we're going to take a pre-created model and replace any torch.nn.ModuleList
+    instances with DistillerModuleLists. Once that is done, during model execution we expect the lists are being
+    as read-only. We're not supporting loading state_dict "across" converted models.
```
nzmora: " Once that is done, during model execution we expect the lists are being
    as read-only."  ==> grammar

distiller/modules/rnn.py
guyjacob: If I remember correctly, @levzlotnik implemented it, and then realized there's a bug in it. And we didn't want to delay the LSTM code, so just commented it out and raised a NotImplementedError. I don't mind that it's there for whenever we want to get back to it.

distiller/summary_graph.py
guyjacob: Fixed, thanks.

distiller/summary_graph.py
guyjacob: Fixed, thanks. I think I wrote it one way, then thought about another way, and somehow got stuck in the middle :)

PR #344, opened 2019-08-06
Refactor Image Classifier Distiller Functions

barrh:
Motivation: I wanted to facilitate the access to the compress_classifier's internal functions, which I use frequently, and potentially, provide different API to same train/valid/test func call.
e.g. "compress_classifier_v2.py"

------------------------------
Move the majority of functionalities from compress_classifier.py to separate file imageclass.py, under distiller.
The function will henceforth will be imported automatically upon calling `import distiller`.

Also, some smaller modifications:
* evaluate_model() calls either test, or quantize and then test, which will now also calibrate.
  Calibration and testing of the quantized models will run on cloned model.
* Rename train() to train_one_epoch()
* Allow calling test and validate/test with less arguments.
  The following arguments are now optional: loggers, collectors, and args.
* acts_quant_stats_collection will return the results stats, and saving
  to file is not optional, false by default.
* qe_calibration argument is set to None, by default.
* Added get_nonparallel_clone_model to handle cloning and discarding of models.

### DIFF
--- a/distiller/__init__.py
+++ b/distiller/__init__.py
@@ -26,6 +26,7 @@ from .policy import *
 from .thinning import *
 from .knowledge_distillation import KnowledgeDistillationPolicy, DistillationLossWeights
 from .summary_graph import SummaryGraph, onnx_name_2_pytorch_name
+import imageclass
 
 import logging
 logging.captureWarnings(True)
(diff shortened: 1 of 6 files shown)


barrh:
Replaced by: https://github.com/NervanaSystems/distiller/pull/346

PR #346, opened 2019-08-07
Enable quantized evaluation in single call

barrh:
In order to apply post-quantization, no pre-made stats file is required.
Calibration is executed implicitly, and its results are discarded after use.

* qe_calibration argument is set to None, by default.
* acts_quant_stats_collection() will return the results stats, and saving
  to file is not optional, false by default.
* evaluate_model() calls either test, or quantize and then test, which will now also calibrate.
  Calibration and testing of the quantized models will run on cloned model.
  Also, API change.
* Added get_nonparallel_clone_model() to handle cloning and discarding of models.


distiller/apputils/image_classifier.py
```diff
@@ -789,7 +790,7 @@ def earlyexit_validate_stats(args):
     return total_top1, total_top5, losses_exits_stats
 
 
-def evaluate_model(model, criterion, test_loader, loggers, activations_collectors, args, scheduler=None):
+def evaluate_model(test_loader, model, criterion, loggers, activations_collectors=None, args=None, scheduler=None):
```
guyjacob: Why switch the order of the parameters?

distiller/apputils/image_classifier.py
```diff
@@ -799,30 +800,50 @@ def evaluate_model(model, criterion, test_loader, loggers, activations_collector
...
+
+
+def quantize_and_test_model(test_loader, model, criterion, loggers=None, activations_collectors=None, args=None,
+                            scheduler=None, save_flag=False):
+    if args is None:
+        raise NotImplementedError('args cannot be set to None')
+
+    if not args.qe_stats_file:
+        args_copy = copy.deepcopy(args)
+        args_copy.qe_calibration = args.qe_calibration if args.qe_calibration is not None else 0.05
```
guyjacob: So far, if a user just specifies `--quantize-eval` with no stats file, then we do dynamic quantization. The way you did it, this functionality is gone. Since I'm not really supporting the dynamic mode, I don't think it's a big deal to change the behavior. But I would ask that you add a new option called `--qe-dynamic`, so if someone really really wants that, it can be invoked

distiller/apputils/image_classifier.py
```diff
@@ -799,30 +800,50 @@ def evaluate_model(model, criterion, test_loader, loggers, activations_collector
...
+        args_copy.effective_test_size = args_copy.qe_calibration
+        args.qe_stats_file = acts_quant_stats_collection(load_data(args_copy)[2], model, criterion, loggers, args_copy)
+
+    with distiller.get_nonparallel_clone_model(model) as qe_model:
```
guyjacob: Why go non-parallel here? There's limitation I'm aware of to quantize and then evaluate a parallel model. Did you notice something like that?

examples/classifier_compression/compress_classifier.py
```diff
@@ -99,7 +99,11 @@ def load_test_data(args):
                                                 args.dataset, add_softmax=True, verbose=False)
         do_exit = True
     elif args.qe_calibration:
```
guyjacob: The way you left it, if someone does `--qe-calibration --quantize-eval` it'll still exit after the calibration. That's not what you meant, was it?
I think it should be something like this, but double check me:

```suggestion
     elif args.qe_calibration and not (args.evaluate and args.quantize_eval):
```

distiller/apputils/image_classifier.py
barrh: uniformity with train, validate and test.
Since the API has changed recently, I thought it would be good timing for that.

distiller/apputils/image_classifier.py
barrh: yeah, I didn't think about it: `--qe-static` and `--qe-dynamic` perhaps?

distiller/apputils/image_classifier.py
barrh: Not at all. The reason here was that I frequently would be cloning to CPU in the context of this commit, and I didn't want to create a specific clone utility for CPU, so I figured that would be the safest method to go about it.

guyjacob: I still don't follow. Then only thing you changed here was forcing quantization to run on a non-parallel model. What's the connection to frequent cloning to CPU? What was wrong with the way it was before?

barrh: Forcing in to be non-parallel in required in order to run parallel source model on CPU.
Henceforth, I will not move the model to the CPU, only when `--qe-cpu` is set.

distiller/apputils/image_classifier.py
```diff
@@ -819,30 +820,51 @@ def evaluate_model(model, criterion, test_loader, loggers, activations_collector
...
+    return test_res
 
-    if args.quantize_eval:
-        checkpoint_name = 'quantized'
-        apputils.save_checkpoint(0, args.arch, model, optimizer=None, scheduler=scheduler,
-                                 name='_'.join([args.name, checkpoint_name]) if args.name else checkpoint_name,
-                                 dir=msglogger.logdir, extras={'quantized_top1': top1})
 
-def acts_quant_stats_collection(model, criterion, loggers, args):
+def acts_quant_stats_collection(test_loader, model, criterion, loggers, args, save_to_file=False):
```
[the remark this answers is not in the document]
barrh: That's mostly for uniformity purposes.

distiller/apputils/image_classifier.py
```diff
@@ -830,30 +844,65 @@ def evaluate_model(model, criterion, test_loader, loggers, activations_collector
...
+    """Collect stats using test_loader, clone the model and quantize the clone, and finally, test it.
 
-    if args.quantize_eval:
+    args.device is allowed to differ from the model's device.
+    When args.qe_calibration is set to None, uses 0.05 instead.
+
+    scheduler - pass scheduler to store it in checkpoint
+    save_flag - defaults to save both quantization statistics and checkpoint.
+    """
+    if not (args.qe_dynamic or args.qe_stats_file):
```
guyjacob: Looks good, thanks for making all the changes!

One last thing - When the `--qe-config-file <path_to_yaml>` arg is used, the contents of the YAML file are taken as-is and all other arguments are ignored. So with this if condition as-is, stats are collected but are ignored when creating the quantizer.
For the sake of merging this PR, I suggest that when a config file is used we skip stats collection. We can consider if we want to change this behavior after. So that means:
```suggestion
    if not (args.qe_dynamic or args.qe_stats_file or args.qe_config_file):
```

Actually - you created a mutually exclusive args group in the quantization args, which currently contains `qe-dynamic` and `qe-stats-file`. I think it should also include `qe-calibration` and `qe-config-file`, since these are all conflicting ways to create stats or use existing ones.

What do you say?

barrh: see new commit

guyjacob:
Merged. Thanks again!

PR #348, opened 2019-08-08
LpRankedStructureParameterPruner: work with 3D filters (ie Conv1d)

trougnouf:
I made these changes to work with the L1RankedStructureParameterPruner_AGP class. The network gets pruned and there are no errors.

The sparsity summary shows 2D/3D pruning only on the 4D filter (screenshot included), I don't know if that's normal.
Screenshot from 2019-08-08 11-16-53

### DIFF
--- a/distiller/thresholding.py
+++ b/distiller/thresholding.py
@@ -85,7 +85,7 @@ def group_threshold_binary_map(param, group_type, threshold, threshold_criteria)
         return binary_map
 
     elif group_type == '3D' or group_type == 'Filters':
-        assert param.dim() == 4, "This thresholding is only supported for 4D weights"
+        assert param.dim() == 4 or param.dim() == 3, "This pruning is only supported for 3D and 4D weights"
         view_filters = param.view(param.size(0), -1)
         thresholds = torch.Tensor([threshold] * param.size(0)).to(param.device)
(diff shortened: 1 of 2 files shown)


distiller/thresholding.py
```diff
@@ -153,7 +153,7 @@ def group_threshold_mask(param, group_type, threshold, threshold_criteria, binar
     elif group_type == '3D' or group_type == 'Filters':
         if binary_map is None:
             binary_map = group_threshold_binary_map(param, group_type, threshold, threshold_criteria)
-        a = binary_map.expand(param.size(1) * param.size(2) * param.size(3), param.size(0)).t()
+        a = binary_map.expand(torch.tensor(param.shape)[1:].prod(), param.size(0)).t()
```
nzmora: I like your idea of using prod() to simplify the multiplication statement.
Syntactically my suggestion below, feels "lighter".  I am biased here, with my own preferences, but I prefer this shorter syntax.  What do you think?
```
np.prod(param.shape[1:])
```

trougnouf: Indeed that is much prettier and there is no reason to convert a list to pytorch tensor. I've committed the np syntax in 0d1d9794ad8f1d40bd5a63681d8a774bc0706b4a , thank you.

PR #355, opened 2019-08-14
Modified fold_batch_norms to allow training loop

levzlotnik:


distiller/quantization/quantizer.py
```diff
@@ -222,6 +222,9 @@ def prepare_model(self, dummy_input=None):
 
         self._pre_process_container(self.model)
         for module_name, module in self.model.named_modules():
+            if getattr(module, 'is_fused', False):
```
guyjacob: Not sure what this is for. I don't see `is_fused` being set anywhere in the code.

distiller/quantization/range_linear.py
```diff
@@ -1109,11 +1110,16 @@ def _apply_bn_folding(self, dummy_input):
         for n, m in named_modules.items():
             try:
                 # Look for the mark left by distiller.model_transforms.fold_batch_norms
-                folded_bn_module = distiller.normalize_module_name(m.fused_modules[0])
+                folded_bn_module = distiller.normalize_module_name(m.fused_sequence[0])
```
guyjacob: What's with the name change here from `fused_modules` to `fused_sequence`? You didn't change the code that actually sets this attribute, there it's still named `fused_modules`:

https://github.com/NervanaSystems/distiller/blob/cb02798ae23079965dbef93ffe4a828f3fa74f8b/distiller/model_transforms.py#L144

As is, when this line is executed it goes straight to the `except` clause below (Which is too broad, so no error is generated. I'll fix that)

distiller/quantization/range_linear.py
```diff
@@ -1109,11 +1110,16 @@ def _apply_bn_folding(self, dummy_input):
...
+
                 # Propagate the output stats of the folded BN module to this module
                 # If stats were collected after folding was applied, then stats for the BN module won't exist,
                 # in which case we just move on
-                model_stats[distiller.normalize_module_name(n)]['output'] = model_stats.pop(folded_bn_module)['output']
+                model_stats[m.fused_module_name]['output'] = model_stats.pop(folded_bn_module)['output']
```
guyjacob: Again, what is this attribute `fused_module_name`? I don't see it being set anywhere. Even if line 1113 would pass (see separate comment on that), this one would cause an exception.

distiller/quantization/range_linear.py
```diff
@@ -1131,7 +1137,8 @@ def _apply_activation_stats_fusions(self):
         named_modules = OrderedDict(self.model.named_modules())
         model_stats = self.model_activation_stats
         for n, m in named_modules.items():
-            if distiller.has_children(m) or n not in self.adjacency_map or len(self.adjacency_map[n].successors) != 1:
+            if (distiller.has_children(m) and not isinstance(m, SimulatedFoldedBatchNorm) )\
```
guyjacob: Is this intended as a bug-fix? Or is it in preparation for use in training?

If the latter, then note that this function is very specific to post-training quantization - updating the collected stats. I don't see it being used in any quant-aware training scenario.

If the former, then note that when we fold batch-norms for inference, we actually return just the param module, not the wrapping `SimulatedFoldedBatchNorm` module: `return folded_module.param_module`
So we shouldn't be running into any `SimulatedFoldedBatchNorm` at this point.

PR #356, opened 2019-08-17
support quantize weights/bias only

hunterkun:
I add support quantizing weights only in the post training quantization when users set `qe-bits-acts=None`. It directly uses quantize-dequantize operations on weights and keep activations fp32 and no warp on layers. In this implementation, i follow @guyjacob 's advice and put them in replace_param_layer of base Quantizer class. When users use config file and set the bits of activation to None, it would be string 'None'  so i give a special handle to this case in quantizer.py.


nzmora:
@levzlotnik @guyjacob - do you want to merge the PR, or should we close it?

guyjacob:
We shouldn't close it. I'll take a look.

guyjacob:
Handled in #439. Thanks @hunterkun and apologies for the long delay.

PR #369, opened 2019-08-29
Clarify usage of Quantizer overrides with DataParallel

levzlotnik:
Added note in `Quantizer` class that overrides layer names should come in their un-`dataparallel`-ized versions.

### DIFF
--- a/distiller/quantization/quantizer.py
+++ b/distiller/quantization/quantizer.py
@@ -156,7 +156,8 @@ class Quantizer(object):
             if isinstance(model, nn.DataParallel):
                 # Need to account for scenario where model is parallelized with DataParallel, which wraps the original
                 # module with a wrapper module called 'module' :)
-                name_to_match = re.sub('module.', '', name_to_match)
+                # i.e. remove only `module.` prefix from everything
+                name_to_match = re.sub('module\\..*', '', name_to_match)
             qbits = self.default_qbits
             override_entry = self.overrides.get(name_to_match, OrderedDict())
             if regex_overrides:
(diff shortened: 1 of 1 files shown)


distiller/quantization/quantizer.py
```diff
@@ -148,9 +152,12 @@ def __init__(self, model, optimizer=None,
...
         for module_full_name, module in model.named_modules():
-            # Need to account for scenario where model is parallelized with DataParallel, which wraps the original
-            # module with a wrapper module called 'module' :)
-            name_to_match = module_full_name.replace('module.', '', 1)
+            name_to_match = module_full_name
+            if isinstance(model, nn.DataParallel):
+                # Need to account for scenario where model is parallelized with DataParallel, which wraps the original
+                # module with a wrapper module called 'module' :)
+                # i.e. remove only `module.` prefix from everything
+                name_to_match = re.sub(r'module\.', '', name_to_match)
```
guyjacob: +1 for checking the actual module type instead of blindly replacing 'module'.

However:
1. Did you mean to check the root model each time? That won't cover cases where the parallelization is only on part of the model. See for example:
https://github.com/NervanaSystems/distiller/blob/2f3a9ac13afe7ad8e1a24c1633c4836c6e28d78a/distiller/models/__init__.py#L115-L118
2. You switched to replacing all instances of `'module.'`. Is that intentional? Not sure which cases you're aiming for with that - Nested DataParallels?

How about enhancing [`utils.normalize_module_name()`](https://github.com/NervanaSystems/distiller/blob/master/distiller/utils.py#L123) to accept the actual model, similar to `denormalize_module_name` below it? And actually look for any DataParallel sub-modules and change the name accordingly. If your intention was indeed to replace only if/when a DataParallel exists, I don't see any way around actually looking for it.

(And as a side note, could have used the same string replace function as the existing code and just remove the last argument (the count argument). No need for regex when a simple string function does the job)

levzlotnik: lol to be honest I just wanted the regex to seek for the `module.` appearance in the beginning of the string, so I forgot to add `^module\.`

PR #387, opened 2019-09-22
Add new flags to activation collector class

barrh:
This patch extends the options on the activation collector class:
1. Allow the option to treat classes argument as black list instead of white list.
2. Allow to collect summary on inputs, not just outputs. Currently mutually exclusive.
3. Print matrix size (useful for weighted averages) [This change reflects on the Excel files]

This is not necessarily final commit, many improvements can be done.
If that's of interest, this patch can be extended to more collector classes.


nzmora:
Closing due to staleness.

PR #399, opened 2019-10-05
Quant-aware training: Better handling of optimizer and creation of float weights copy

guyjacob:
* Create float copy such that the actual tensor being learned stays the same
* This way the optimizer doesn't have to be re-created, just need to add parameter groups if algo requires it (e.g. PACT)
* This also means we don't care about pre-existing parameter groups, as opposed to the previous implementation which **assumed** a single existing group


guyjacob:
Tests passing

PR #406, opened 2019-10-15
Set validation split to 0 by default (BREAKING)

barrh:
@nzmora what do you think?


barrh:
Branch was rebased to avoid conflict in one of the examples file. No code changes.

barrh:
This change doesn't benefit users of CIFAR10 dataset.

PR #410, opened 2019-10-27
Correct handling of different batch-sizes in collectors

levzlotnik:
Pointed out by @Pavelrst :)
Standard deviation calculation was calculated with an average over batches, where the last batch might be very small and hence less elements.  
Corrected by calculating std in the second pass by definition (after `mean` was calculated) .

### DIFF
--- a/distiller/data_loggers/collector.py
+++ b/distiller/data_loggers/collector.py
@@ -479,10 +479,9 @@ class QuantCalibrationStatsCollector(ActivationStatsCollector):
                 prev_mean (float): the previous running mean
                 total_values_so_far (int): the number of the values so far
             """
-            curr_mean = values.mean().item()
             curr_numel = values.numel()
             prev_numel = total_values_so_far
-            return (prev_numel * prev_mean + curr_numel * curr_mean) / (prev_numel + curr_numel)
+            return (prev_numel * prev_mean + values.sum().item()) / (prev_numel + curr_numel)
 
         def update_std(values, prev_std, mean, total_values_so_far):
             """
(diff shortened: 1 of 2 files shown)


barrh:
Please note that the issue with the weighting of the last batch is not exclusive to this class, similar issue arises in SummaryActivationStatsCollector (and its derivatives). So introducing this correction into master, without adding the nuance of the last batch equally across collectors is something to consider.

Another consideration is that these code additions quite inflates the code, reducing its readability, for a functionality that is already found in [AverageValueMeter](https://tnt.readthedocs.io/en/latest/_modules/torchnet/meter/averagevaluemeter.html#AverageValueMeter). While I have my reservations of torchnet's AverageValueMeter implementation (I humbly confess I don't understand the conditions on 'n'), I think a generator object that keeps track of mean/std/var is the way to go.

levzlotnik:
@barrh `AverageValueMeter` is actually implemented in the way I am aiming to correct, i.e. the implementation is very similar to what was before this PR:
https://github.com/NervanaSystems/distiller/blob/61dfc8284e946ccd166792ab4468f3c08765ce81/distiller/data_loggers/collector.py#L444-L452

vs 

https://github.com/pytorch/tnt/blob/d0d7fc966482cfe0e93283697045f07610a6c5e3/torchnet/meter/averagevaluemeter.py#L26-L29 

In fact - to make their version correct it should have been (according to Welford's weighted algorithm):
```
self.m_s += n*(value - self.mean_old) * (value - self.mean)
```

I added this simple second pass algorithm since the second pass for quant_stats was already there for the `b` parameter.

About `SummaryActivationStatsCollector` - `AverageValueMeter` is already in there.

barrh:
@levzlotnik 
The major benefit of AverageValueMeter over your approach is that it sits in a separate module, thus, easier to evaluate & validate, and later reuse in the repo - outside QuantCalibrationStatsCollector. It also has a minor benefit, its API requires only single call to write the new values.

Just to be 100% clear, I don't think one should use torchnet's AverageValueMeter to compute weighted average means, regardless, the concept of a separate component is superior for the reasons above.

Issue #413, opened 2019-10-31
ResneXt50 (torchvision) quantization low accuracy


barrh:
Running the following produces very low accuracy rate (top1 < 1%):
```
python3 examples/classifier_compression/compress_classifier.py --arch=resnext50_32x4d --pretrained IMAGENET --vs=0 --qe-calibration=0.05
python3 examples/classifier_compression/compress_classifier.py --arch=resnext50_32x4d --pretrained IMAGENET  --vs=0 --qe-clip-acts=avg --qe-no-clip-layers=fc --qe-mode=asym_u --eval --quantize-eval --qe-stats-file=[...]/acts_quantization_stats.yaml
```

Am I doing something wrong?

guyjacob:
See #414. I tested it on `resnext50_32x4d` and results are sensible now.

barrh:
Resolved in master.

PR #414, opened 2019-10-31
Update internal ResNet implementation according to latest TorchVision

guyjacob:
Changed our version to only re-implement BasicBlock and Bottleneck,
and duplicate the model creation functions. Other than that, re-use
everything from the torchvision implementation.


barrh:
I tested this commit on ResNeXt50_32x4d merged with PR #346 , as described in issue #413 , and it works as expected.

PR #420, opened 2019-11-06
Deepcopy Namespace argument in Classifier class init function

barrh:
This fixes a bug where classifier_obj.args will being modified by the caller inadvertently.


guyjacob:
Fixed in #421 - Thanks @barrh!

Issue #411, opened 2019-10-28
Quantization aware training


jonathanbonnard:
Hi,
I 'm training a mobilenetv2 on cifar from scratch using 8 bits on weights + activations.(QuantAwareTrainRangeLinearQuantizer)

> --- validate (epoch=64)-----------
10000 samples (512 per mini-batch)
Epoch: [64][   10/   19]    Loss 0.797462    Top1 73.007812    Top5 97.988281
Epoch: [64][   20/   19]    Loss 0.793207    Top1 72.870000    Top5 97.890000
==> Top1: 72.870    Top5: 97.890    Loss: 0.793

`==> Best [Top1: 58.000   Top5: 95.020   Sparsity:0.45   Params: -4419393 on epoch: 12]
Saving checkpoint to: logs/2019.10.28-114149/checkpoint.pth.tar`

As you can see the validation step gives me a 72% top 1 accuracy, but it keeps a 58% top 1 as best checkpoint. 
Is this an issue or something I didn't understand?

jonathanbonnard:
ok, it turns out that the validation step keeps the model with the highest sparsity level. That's not what I expected when using a quantizer

nzmora:
Hi @jonathanbonnard,

Yes, you are correct.  You can see the sorting code in this line: https://github.com/NervanaSystems/distiller/blob/c849a25fd8d8719c1a81d43f85cf831d24f6da37/distiller/apputils/image_classifier.py#L736

This line first sorts the results by `params_nnz_cnt`, followed by `top1`, `top5`, and finally `epoch`.  So the most important ordering feature is `params_nnz_cnt`, which is the count of non-zero (nnz) parameters. This was written from the point-of-view of pruning, and not quantization, and I didn't expect this to influence quantization. However, because there are some zero-coefficients in epoch 12 (in your example above), this sorting scheme breaks the expected behavior.

You can of course change this line to get a different ordering.  
You can also issue a PR with a fix (e.g. when quantization is involved, use a different sorting criteria).
Thanks for pointing out this issue!
Neta

jonathanbonnard:
Thank you @nzmora, this solved my issue.

PR #427, opened 2019-11-14
Perf tracker

nzmora:
Introduce a performance tracking class

This will help define and use different performance sorting schemes.
E.g. this will address the issue raised in issue #411

### DIFF
--- a/distiller/apputils/image_classifier.py
+++ b/distiller/apputils/image_classifier.py
@@ -157,7 +157,7 @@ class ClassifierCompressor(object):
 
     def _finalize_epoch(self, epoch, top1, top5):
         # Update the list of top scores achieved so far, and save the checkpoint
-        self.performance_tracker.step(self.model, top1, top5, epoch)
+        self.performance_tracker.step(self.model, epoch, top1=top1, top5=top5)
         _log_best_scores(self.performance_tracker, msglogger)
         best_score = self.performance_tracker.best_scores()[0]
         is_best = epoch == best_score.epoch
--- a/distiller/apputils/image_classifier.py
+++ b/distiller/apputils/image_classifier.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2018 Intel Corporation
+# Copyright (c) 2019 Intel Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -28,16 +28,15 @@ import torch.backends.cudnn as cudnn
 import torch.optim
 import torch.utils.data
 import torchnet.meter as tnt
+import parser
+from functools import partial
+import argparse
 import distiller
 import distiller.apputils as apputils
 from distiller.data_loggers import *
 import distiller.quantization as quantization
 import distiller.models as models
 from distiller.models import create_model
-import parser
-from functools import partial
-import operator
-import argparse
 from distiller.utils import float_range_argparse_checker as float_range
 
 # Logger handle
@@ -76,6 +75,7 @@ class ClassifierCompressor(object):
         self.train_loader, self.val_loader, self.test_loader = (None, None, None)
         self.activations_collectors = create_activation_stats_collectors(
             self.model, *self.args.activation_stats)
+        self.performance_tracker = apputils.SparsityAccuracyTracker(self.args.num_best_scores)
     
     def load_datasets(self):
         """Load the datasets"""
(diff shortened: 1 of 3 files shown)


distiller/apputils/performance_tracker.py
```diff
@@ -0,0 +1,72 @@
...
+# limitations under the License.
+#
+
+"""Performance trackers used to track the best performing epochs when training.
+"""
+import operator
+import distiller
+
+
+__all__ = ["SparsityAccuracyTracker"]
```
barrh: Can you expose both?
I would prefer customizing in classifier.py than here.

nzmora: done; But the reason I didn't expose it in the first place is that I think that we can all benefit from sub-classes (for handling ranking based on number of quantized weights and activations) - so think of that when you decide where to place your sub-classes.

distiller/apputils/performance_tracker.py
```diff
@@ -0,0 +1,65 @@
...
+class AccuracyPerformanceTracker(object):
+    """Base class for performance trackers using Top1 and Top5 accuracy metrics"""
+    def __init__(self, num_best_scores):
+        self.perf_scores_history = []
+        self.max_len = num_best_scores
+
+    def reset(self):
+        self.perf_scores_history = []
+
+    def step(self, model, top1, top5, epoch):
```
guyjacob: Maybe use a list of score-names and score-values pairs instead of hard coded top-1/5? I mean, one could always "abuse" this in non-classification tasks and pass different metrics, but that's not intuitive.

distiller/apputils/performance_tracker.py
```diff
@@ -0,0 +1,65 @@
...
+        """Update the list of top training scores achieved so far"""
+        raise NotImplementedError
```
guyjacob: Why not have a default implementation using just accuracy values?

distiller/apputils/image_classifier.py
```diff
@@ -76,6 +75,7 @@ def __init__(self, args, script_dir):
         self.train_loader, self.val_loader, self.test_loader = (None, None, None)
         self.activations_collectors = create_activation_stats_collectors(
             self.model, *self.args.activation_stats)
+        self.performance_tracker = apputils.SparsityAccuracyTracker(self.args.num_best_scores)
```
guyjacob: I don't think this should be the default - you'd get the same problem as #411. Assuming that was the motivation for this, maybe defer the selection of the tracker until a compression scheduler was created, and only if pruning/regularization policies exist then use this tracker. This also assumes a default tracker that only looks at accuracy (see comment below).
(I realize that the way you implemented it current functionality - so this suggestion can be deferred to a later PR if you want)

distiller/apputils/performance_tracker.py
barrh: @guyjacob I agree that enabling flexible API can benefit the user, but in this case, I feel that it has the potential to over-complicate the API for limited benefit.

e.g. If I were to come up with alternative params for ranking it would be: `(top1, loss, epoch)`
- top1 - as before
- loss - Distinguished from top5 due to lower-is-better attribute, and has to scope of the entire universe of classes.
- epoch - for stable ordering

To allow to above params, the "pairs" API should first be upgraded to "trios" API: (name,value,asc/desc).
For conclusion, pairs don't benefit me, but trios do, then again - complicated.

distiller/apputils/performance_tracker.py
nzmora: @guyjacob - please review the new API.  It is now more generic.

distiller/apputils/performance_tracker.py
nzmora: @guyjacob, @barrh  - please review the new API.  It is now more generic.
@barrh - to control the "direction" of the ranking (what you refer to as asc/desc above) just pass the score multiplied by -1.
For example, in `SparsityAccuracyTracker` I want to prefer lower parameter NNZ count, so I sort by `-params_nnz_cnt`.

barrh: @nzmora Wouldn't it reflect in the logging as negative value?

nzmora: @barrh the logging code is orthogonal and handles this:
`self.performance_tracker.step(self.model, epoch, top1=top1, top5=top5)
        _log_best_scores(self.performance_tracker, msglogger)
`

`_log_best_scores` is a private function of `image_classifier.py` and is currently specific to `apputils.SparsityAccuracyTracker` (I just pushed a commit to make this explicit in code).  If/when someone uses a different PerformanceTracker, they can generalize this logging function.

PR #428, opened 2019-11-17
Added optional requires_grad to linear quantization parameters.

levzlotnik:
For future use in automatic quantization search


guyjacob:
Implemented in #432

PR #432, opened 2019-11-27
Loss Aware Post Train Quantization search

levzlotnik:
as per https://arxiv.org/abs/1911.07190 and their github repo.  
Many thanks to @ynahshan et al


levzlotnik:
example command line -
```time python3 ptq_coordinate_search.py -a resnet50 --pretrained -j 22 /data/datasets/imagenet --opt-test-size 0.01 --opt-maxiter 2 --qe-stats-file resnet50_act_stats.yaml --qe-no-quant-layers conv fc -b 500 --gpu 0 --search-on scale --opt-init-mode L3 --opt-eval-single-batch```

**Comment by levzlotnik** [comment]
Updated args:
```
time python3 ptq_coordinate_search.py -a resnet50 --pretrained -j 22 /data/datasets/imagenet --opt-test-size 0.01 --opt-maxiter 2 --qe-stats-file resnet50_act_stats.yaml --qe-no-quant-layers conv fc -b 500 --gpu 0 --search-on scale --opt-init-mode L3 --opt-eval-memoize-dataloader
```
`--opt-eval-memoize-dataloader` stores all batches in memory, for faster performance (no need to call for pre-processing).

File: distiller/quantization/ptq_coordinate_search.py [hunk 1]
```diff
@@ -0,0 +1,424 @@
...
+    args.effective_test_size = args.qe_calib_portion
+    args.batch_size = args.qe_calib_batchsize
+    calib_data_loader = classifier.load_data(args, fixed_subset=True, load_train=False, load_val=False)
+    # logging
+    logging.getLogger().setLevel(logging.WARNING)
+    msglogger = logging.getLogger(__name__)
+    msglogger.setLevel(logging.INFO)
+    model = create_model(args.pretrained, args.dataset, args.arch,
+                         parallel=not args.load_serialized, device_ids=args.gpus)
+    device = next(model.parameters()).device
```
nzmora: using `distiller.utils.model_device` is preferred because: (1) it explains the intent better; (2) it hides the implementation details.  In this specific case (2) is important because this implementation is only partially true (it is correct if the entire model resides on a single device).  If we every make `distiller.utils.model_device` more robust, then this app will immediately benefit.

levzlotnik:
Update args:
```
time python3 ptq_coordinate_search.p -a resnet50 --pretrained -j 22 /data/datasets/imagenet --opt-val-size 3.5e-3 --opt-maxiter 2 --qe-stats-file resnet50_act_stats.yaml --qe-no-quant-layers conv fc -b 500 --search-on scale --opt-init-mode L3 --opt-eval-memoize-dataloader
```

guyjacob:
Not done yet:
* Proper documentation
* Interpolation of optimal LP value for initialization
* Refactor the "script" part (basically the `__main__` function) to a separate script under the examples.

**Differences in initialization phase:**
* In the reference implementation:
  * When doing LP initialization, weights and activations are initialized based on quantization loss of their respective tensors.
  * Activations are initialized "online", meaning the input to the layer N being initialized is the output of the **already quantized layer N-1**.
* In this implementation:
  * When doing LP initialization for a given layer, we initialize both activations and weights together (as applicable) based on the LP loss between the quantized layer output and the FP32 layer output.

guyjacob:
See an example YAML file:
https://github.com/NervanaSystems/distiller/blob/master/examples/quantization/post_train_quant/resnet18_imagenet_post_train_lapq.yaml

PR #437, opened 2019-12-03
fix error about exporting stats file of quant

Emily0219:
With classifier.acts_quant_stats_collection(), no stats file is saved default.


guyjacob:
Thanks for catching this!

PR #443, opened 2019-12-12
IFM sparsity collector

barrh:
Patch adds support to collecting IFM sparsity, as well as multiple IFMs. If multiple IFMs are present, they are merged with `torch.cat`, which is a parameter to the collector.
IFM sparsity will be collected whenever activation sparsity is collected with `create_activation_stats_collectors`.
Renamed some of the fields/filenames.

Tested against Python3.6. Passed ~~excluding the post-training tests.~~ (merged @guyjacob bugfix)
Note the use of typing library for type hinting. Not sure about the state of support for this in older Python 3.5 environments. It's possible that this patch requires Python3.6 OR Python3.5+updates, but not sure how to check that.

### DIFF
--- a/distiller/data_loggers/collector.py
+++ b/distiller/data_loggers/collector.py
@@ -18,6 +18,7 @@ import contextlib
 from functools import partial, reduce
 import operator
 import xlsxwriter
+import enum
 import yaml
 import os
 from sys import float_info
@@ -38,8 +39,16 @@ msglogger = logging.getLogger()
 
 __all__ = ['SummaryActivationStatsCollector', 'RecordsActivationStatsCollector',
            'QuantCalibrationStatsCollector', 'ActivationHistogramsCollector',
+           'CollectorDirection',
            'collect_quant_stats', 'collect_histograms',
            'collector_context', 'collectors_context']
+
+class CollectorDirection(enum.Enum):
+    OUT = 0
+    OFM = 0
+    IN = 1
+    IFM = 1
+    IFMS = 1
 
 
 class ActivationStatsCollector(object):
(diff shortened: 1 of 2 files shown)


barrh:
@nzmora I would like to replace `lambda t: 100 * distiller.utils.sparsity(t)` with `distiller.utils.sparsity`.
It make sense to me because when you open it in Excel, this should be formatted in percentage, which  achieves the intended result, whereas now it does not. e.g. reformatting "20.5" to percentage in Excel results in "2050%".
What do you think?

distiller/data_loggers/collector.py
```diff
@@ -251,28 +260,28 @@ def _collect_activations_stats(self, module, activation_stats, name=''):
             activation_stats[getattr(module, self.stat_name).name] = mean
 
     def save(self, fname):
```
nzmora: why did you remove the remarks?

distiller/data_loggers/collector.py
```diff
@@ -201,20 +204,27 @@ class SummaryActivationStatsCollector(ActivationStatsCollector):
...
+    output_flag - When True collect on OFM, otherwise, IFMs
+    inputs_consolidate_func is called on tuple of tensors, and returns a tensor.
     """
-    def __init__(self, model, stat_name, summary_fn, classes=[torch.nn.ReLU,
-                                                              torch.nn.ReLU6,
-                                                              torch.nn.LeakyReLU]):
+    def __init__(self, model, stat_name, summary_fn,
+                 classes=[torch.nn.ReLU, torch.nn.ReLU6, torch.nn.LeakyReLU],
+                 output_flag=True,
+                 inputs_consolidate_func:typing.Callable[[typing.Sequence[TensorType]], TensorType]=torch.cat):
```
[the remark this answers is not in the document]
barrh: Since it's not critical for the logic of the code, I could do without it. Removing this code has the side-benefit of undoing the dependency on typing library integration, which is introduced in this commit.
However, I think that enhancing documentation of complex arguments (e.g. callables) is a blessing.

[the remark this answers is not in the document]
nzmora: i agree about the benefit of enhanced documentation - but I think that if we make a change in the code style, it should be more organized

distiller/data_loggers/collector.py
barrh: It was incorrect.

distiller/data_loggers/collector.py
```diff
@@ -203,23 +212,24 @@ class SummaryActivationStatsCollector(ActivationStatsCollector):
     light-weight and quicker than collecting a record per activation.
     The statistic function is configured in the constructor.
 
-    output_flag - When True collect on OFM, otherwise, IFMs
+    collector_direction - enum type: IN for IFMs, OUT for IFM
```
nzmora: OUT for **O**FM

distiller/data_loggers/collector.py
```diff
@@ -38,9 +39,17 @@
 
 __all__ = ['SummaryActivationStatsCollector', 'RecordsActivationStatsCollector',
            'QuantCalibrationStatsCollector', 'ActivationHistogramsCollector',
+           'CollectorDirection',
            'collect_quant_stats', 'collect_histograms',
            'collector_context', 'collectors_context']
 
+class CollectorDirection(enum.Enum):
```
nzmora: can you add a documentation line explaining what IN/OUT and IFM/OFM (input/output feature maps) are?

PR #518, opened 2020-06-17
WIP: Upstream torch1.5.0

cgerum:
This is an initial proposal for a version that should also be compatible with pytorch 1.4.0 and 1.5.0.

It has not yet been tested extensively, but the unit tests pass.

### DIFF
--- a/distiller/summary_graph.py
+++ b/distiller/summary_graph.py
@@ -22,10 +22,50 @@ import torch
 import torch.nn as nn
 import torch.jit as jit
 import logging
+from pkg_resources import parse_version
 from collections import OrderedDict, defaultdict
 from collections.abc import MutableSequence, Iterable
 msglogger = logging.getLogger()
 
+# see: https://github.com/pytorch/pytorch/issues/33463#issuecomment-606399944
+class scope_name_workaround(object):
+    def __init__(self):
+        self.backup = None
+
+    def __enter__(self):
+        def _tracing_name(self_, tracing_state):
+            if not tracing_state._traced_module_stack:
+                return None
+            module = tracing_state._traced_module_stack[-1]
+            for name, child in module.named_children():
+                if child is self_:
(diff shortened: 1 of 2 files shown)


rotx-maxim:
@cgerum, thanks for this patch! I found I also have to do a torch.onnx.symbolic_helper._set_opset_version(11).
We have a few other changes for PyTorch 1.5, see https://github.com/MaximIntegratedAI/distiller, mainly a new TensorBoard backend.

cgerum:
Thanks, I hope to have some time next week to work on the 1.5.0 compatibility again. 
Do you know which layer were problematic without _set_opset_version(11).

rotx-maxim:
The one I remember is round(), but there may have been others.

TigerVersusT:
Cool, @cgerum , does it fully support all commands in original repository? I tried another  [fork](https://github.com/MaximIntegratedAI/distiller) but the 1.7 branch seems doesn't support "--export-to-onnx" and "--thinify"

cgerum:
@TigerVersusT Sorry I suppose, that those commands will not work with this patch set. As distiller is pretty dead for over 2 years, I would recommend migrating to another minimization framework. 

Some good ones are:
- Xilinix brevitas for quantization: https://github.com/Xilinx/brevitas
- Qualcomm AIMET for quantization and pruning based compression techniques: https://github.com/quic/aimet

TigerVersusT:
> @TigerVersusT Sorry I suppose, that those commands will not work with this patch set. As distiller is pretty dead for over 2 years, I would recommend migrating to another minimization framework.
> 
> Some good ones are:
> 
>     * Xilinix brevitas for quantization: https://github.com/Xilinx/brevitas
> 
>     * Qualcomm AIMET for quantization and pruning based compression techniques: https://github.com/quic/aimet

Thanks for your information, that will help me alot.

PR #535, opened 2020-09-08
Resume from checkpoint support for Quantization Aware Training

Akimoto-Cris:
Add Gradients Quantization support for DorefaQuantizer.


Akimoto-Cris:
Also implement gradient quantization for Dorefa

